PR #3, opened 2015-08-21
fixed buffer overflow in qtewin

zerosum0x0:
The source file qtewin.cpp uses strcpy() directly from argv[] input into a 64 byte buffer. This adds a check that will ignore input larger than the buffer. You can also reject to start the program instead, depending on preference.


hean01-cendio:
Please create a new PR without commits adding and removing the temporary qtewin.cpp~

PR #6, opened 2015-08-21
fixed buffer overflow in qtewin.cpp, needed >=

zerosum0x0:
qtewin.cpp reads argv[] directly into a strcpy on a  64 byte buffer


hean01-cendio:
Due to no response to the change and the fact that uiports is to be removed issue #26 I'm closing this  PR.

PR #20, opened 2015-12-08
Allow workspaces to be switched

n-a-m-e:
theres a bunch of people who would like to be able to switch workspaces while using rdesktop
https://bugs.launchpad.net/ubuntu/+source/rdesktop/+bug/861789
http://ubuntuforums.org/showthread.php?t=954887

this patch was originally proposed in 2009 but never got implemented
http://sourceforge.net/p/rdesktop/patches/178/

i think it's really useful ...

originally done by sunner for version rdesktop-1.6.0
http://blog.sunner.cn/rdesktop-patch-switch-local-workspaces-and-windows-by-local-hot-keys/

Double press Ctrl+Alt+Left or Ctrl+Alt+Right to switch local active workspaces.
Double press Ctrl+Alt+Tab to switch local active window.
Other hot keys are sent directly to remote machine.
If using -fD options and having two or above workspaces, the effect is amazing!
If rdesktop is running in fullscreen mode (-f and without -D), Ctrl+Alt+Left/Right/Tab will toggle it to window mode first. After switching back, it can NOT toggle to fullscreen automatically. Press Ctrl+Alt+Enter to do that. (I know this is boring. But it is the best I can do. Rdesktop use override_redirect to implement fullscreen which makes it always the top-most window no matter which workspace/window you have switched to)

### DIFF
--- a/xkeymap.c
+++ b/xkeymap.c
@@ -636,6 +636,24 @@ handle_special_keys(uint32 keysym, unsigned int state, uint32 ev_time, RD_BOOL p
 {
 	switch (keysym)
 	{
+		case XK_Left:
+		case XK_Right:
+		case XK_Tab:
+			if ((get_key_state(state, XK_Alt_L) || get_key_state(state, XK_Alt_R))
+				&& (get_key_state(state, XK_Control_L)
+				|| get_key_state(state, XK_Control_R)))
+			{
+				/* Ctrl-Alt-Left/Right/Tab:
+				* Ungrab the keyboard so that user can use Windows manager's hot keys */
+				extern RD_BOOL g_fullscreen;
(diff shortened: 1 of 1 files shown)


samodadela:
Correction: up/down cursor works - so patch is usable (by pressing ctrl-alt-arrow twice: 1st is for ungrab, 2nd is the ctrl-alt-arrow for the window manager). My problem with up/down was that I just copied the binary to /usr/local/bin without doing a make install - this way rdesktop could not find the keymap files (it said it, also - but I didn't notice) and so most keys did not work. 

Sorry for the confusion.

hean01-cendio:
Closing this PR due to it is a diversion from the solution for issue #22 which would solve the problem to control grab of keyboard to session or not.

PR #23, opened 2016-01-07
Keepalive (mouse wiggler) feature

gsomlo:
RDP servers are often configured to auto-logout on user inactivity, sometimes killing useful long-running but non-interactive programs in the process. This problem is bad enough, to the point where hardware "mouse wiggler" dongles are available for sale (e.g. http://www.amazon.com/CRU-Inc-30200-0100-0011-WiebeTech-Jiggler/dp/B000O3S0PK).

This pull request contains a rebased and slightly cleaned-up version of an older subbmission by Aaron Suen (http://sourceforge.net/p/rdesktop/patches/113/), which implements the mouse wiggler feature in software. This offers a huge advantage over a hardware mouse wiggler, as it limits the "wiggling" to the remote desktop session, allowing my outer Linux desktop to still go to screensaver but keeping the RDP session alive as desired.

The mouse-wiggle/keepalive patch is preceded by factoring out event wallclock time acquisition, cleaning up the event loop slightly, and making the "mouse wiggler" easier and cleaner to implement.

Thanks !

### DIFF
--- a/doc/rdesktop.1
+++ b/doc/rdesktop.1
@@ -123,6 +123,12 @@ but everything after this is unencrypted (including interactive logins).
 Do not send mouse motion events.  This saves bandwidth, although some Windows
 applications may rely on receiving mouse motion.
 .TP
+.TP
+.BR "-M <interval>"
+Send keepalive mouse motion events every <interval> seconds.  This prevents
+the server from timing out the session due to inactivity, and in the process
+terminating potentially important non-interactive programs.
+.TP
 .BR "-C"
 Use private colourmap.  This will improve colour accuracy on an 8-bit display,
 but rdesktop will appear in false colour when not focused.
--- a/rdesktop.c
+++ b/rdesktop.c
@@ -154,6 +154,8 @@ extern RDPDR_DEVICE g_rdpdr_device[];
 extern uint32 g_num_devices;
 extern char *g_rdpdr_clientname;
 
+extern uint32 keepalive_interval;
+
 #ifdef RDP2VNC
 extern int rfb_port;
 extern int defer_time;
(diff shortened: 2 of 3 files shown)


hean01-cendio:
The "problem" your patch solves is to work around a RDS policy set by system administrators, which is not in the scope of rdesktop as a client of RDS.

The best solution to this problem is to make the administrators of the RDS envrionment aware about the problem with long-running, non-interactive process in their environment.

gsomlo:
Henrik,

Thanks for the quick reply !

On Thu, Feb 04, 2016 at 06:25:52AM -0800, Henrik Andersson wrote:
> The "problem" your patch solves is to work around a RDS policy set by
> system administrators, which is not in the scope of rdesktop as a client
> of RDS.

Disappointing, but understandable...

> The best solution to this problem is to make the administrators of the
> RDS envrionment aware about the problem with long-running, non-interactive
> process in their environment.

I have to disagree on this point: if that were true, hardware mouse
wigglers wouldn't be such a popular hot-selling item on e.g. Amazon ;)

Sometimes server admins do what they have to, and users do what little
is in their power, and never the two shall have to fight over it :)

Thanks anyway for looking into it !

--Gabriel

Warr1024:
Hardware dongles are not really a useful option when you are using the client desktop for more than just an RDP client, i.e. the remote session times out because it was in background or minimized.

There is a software workaround for this, but it's quite cumbersome:
- Run `Xephyr` to create an embedded X11 server.
- Run `rdesktop` full-screen within this server.
- Use `xdotool` to check mouse location and "wiggle" it if it has been in the same position for a certain amount of time.
Of course, to do this, I needed a couple hundred lines of perl to hold it all together, and it's a bit fragile.  `Xephyr`, in particular, imposes some performance costs, and has a bad habit of grabbing the cursor on `CTRL`+`Shift` (which apparently cannot be disabled without hacking on the src: http://superuser.com/a/281650).

PR #34, opened 2016-04-07
Factor out event wallclock time acquisition

gsomlo:
Instead of calling 'time(NULL)' from multiple places across the
event loop, compute 'ev_time' once for each event before handling
the event type. Also, pass 'ev_time' into 'handle_button_event()'
for the same reason.
This would make for a slightly cleaner and more efficient event
loop.
---
PS. I've given up on trying to push a mouse-wiggler feature (for which this patch was
a precursor cleanup step), but thinking about it, this should be a genuinely positive
change for rdesktop in and of itself, orthogonally to the previous rejected pull request :)

Please consider applying just this individual patch. Thanks,
--Gabriel

### DIFF
--- a/doc/rdesktop.1
+++ b/doc/rdesktop.1
@@ -123,6 +123,12 @@ but everything after this is unencrypted (including interactive logins).
 Do not send mouse motion events.  This saves bandwidth, although some Windows
 applications may rely on receiving mouse motion.
 .TP
+.TP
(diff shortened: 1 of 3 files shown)


gsomlo:
sorry, didn't realize that github would automatically add new commits to open pull requests. I'll nuke my repo and generate a new one with just the cleanup patch :(
Sorry for the noise...

PR #35, opened 2016-04-07
Factor out event wallclock time acquisition

gsomlo:
Instead of calling 'time(NULL)' from multiple places across the
event loop, compute 'ev_time' once for each event before handling
the event type. Also, pass 'ev_time' into 'handle_button_event()'
for the same reason.
This would make for a slightly cleaner and more efficient event
loop.

----
second attempt at submitting just the cleanup/optimization step of my earlier
pull request. Giving up on upstreaming the mouse wiggler, but hoping just
this patch is of value to rdesktop in and of itself.

Thanks again,
--Gabriel

### DIFF
--- a/xwin.c
+++ b/xwin.c
@@ -2245,7 +2245,7 @@ xwin_toggle_fullscreen(void)
 }
 
 static void
-handle_button_event(XEvent xevent, RD_BOOL down)
+handle_button_event(XEvent xevent, uint32 ev_time, RD_BOOL down)
 {
 	uint16 button, flags = 0;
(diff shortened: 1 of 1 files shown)


hean01-cendio:
Just looked through the patch and i can't agree on that it would improve the efficiency of the event loop. In the current implementation there are 20 event handled where 3 of them makes a call to time(). Your changes makes the call for any event processed and unprocessed. If time() was always used for any event then I would merge your request but IMHO the changes deviates from SPOT and efficiency.

PR #44, opened 2016-06-16
Fix OpenSSL 1.1 compability issues

jjhoo:
Some data types have been made opaque in OpenSSL version 1.1 so
stack allocation and accessing struct fields directly does not work.


hean01-cendio:
I have looked at this PR which looks fine. However I'm concerned about backward compatibility with  OpenSSL 1.0.1 and 1.0.2 line. OpenSSL 1.1 is beta and shouldn't be used in production for now.

hean01-cendio:
I have done some minor testing for backward compatibility. Seems good. Merging this PR now when opeenssl 1.1.0 is released.

PR #56, opened 2016-08-24
Added -M option for left-handed mouse buttons

yu210148:
addresses Issue 15

Adapted from steelman's patch here:
https://sourceforge.net/p/rdesktop/patches/155/


hean01-cendio:
Closing this PR due to this is considered to be a session side configuration. Problems such as MSTSC and rdesktop behaves in different ways when user adds -M option for "left hand support".

Issue #65, opened 2016-10-01
Windows 10 Support


joshudson:
Windows 10 support is currently broken in some cases.

1) DPI must now be sent from the client or the server assumes 96 DPI and won't accept either the registry override or the local monitor size anymore.

2) Remote mouse can get stuck with a bad mouse cursor (issue also exists in Windows Server 2012 R2). Added -M to ignore remote mouse cursor.
##

joshudson:
diff -ru stock/rdesktop-1.8.3/doc/rdesktop.1 work/rdesktop-1.8.3/doc/rdesktop.1
--- stock/rdesktop-1.8.3/doc/rdesktop.1	2016-10-01 12:21:55.000000000 -0700
+++ work/rdesktop-1.8.3/doc/rdesktop.1	2016-10-01 13:47:33.913176792 -0700
@@ -63,15 +63,18 @@
 so its use is discouraged.
 .TP
 .BR "-g <geometry>"
-Desktop geometry (WxH). If geometry is the special word "workarea", the geometry
-will be fetched from the extended window manager hints property _NET_WORKAREA, from
-the root window. The geometry can also be specified as a percentage of the whole
-screen, e.g. "-g 80%". 
-
+Desktop geometry (WxH[@DPI][+X[+Y]]). If geometry is the special word
+"workarea", the geometry will be fetched from the extended window
+manager hints property _NET_WORKAREA, from the root window. The geometry
+can also be specified as a percentage of the whole screen, e.g. "-g 80%". 
 If the specified geometry depends on the screen size, and the screen
 size is changed, rdesktop will automatically reconnect using the new
 screen size. This requires that rdesktop has been compiled with RandR
 support.
+
+The optional DPI parameter should be specified if the screen rdesktop
+is being displayed on is too far from 96 DPI for unscaled Windows to
+be readable. Windows currently accepts values from 96 to 480.
 .TP
 .BR "-i"
 Use password as smartcard pin. If a valid user certificate is matched in smart card
@@ -123,6 +126,11 @@
 Do not send mouse motion events.  This saves bandwidth, although some Windows
 applications may rely on receiving mouse motion.
 .TP
+.BR "-M"
+Use the local X cursor rather than the one from the remote machine. This is most
+useful with "-m" but has other uses. If running with no window manager, use
+xsetroot(1) to set a useful cursor first.
+.TP
 .BR "-C"
 Use private colourmap.  This will improve colour accuracy on an 8-bit display,
[...]

PR #66, opened 2016-10-04
Add Windows 10 support: add option to send DPI to server as Windows 1…

joshudson:
…0 lost the ability to remember remote desktop DPI and add ability to ignore remote mouse cursor as Windows 10 sometimes sends a dumb cursor (not auto-enabled as we can't really detect when we need to do this)
#65


secure.c
```diff
@@ -901,7 +912,7 @@ sec_connect(char *server, char *username, char *domain, char *password, RD_BOOL
 		return False;
 
 	/* We exchange some RDP data during the MCS-Connect */
-	mcs_data.size = 512;
+	mcs_data.size = 542;
```
hean01-cendio: Is this a relevant change for adding DPI scaling support ?

doc/rdesktop.1
```diff
@@ -123,6 +128,11 @@ but everything after this is unencrypted (including interactive logins).
 Do not send mouse motion events.  This saves bandwidth, although some Windows
 applications may rely on receiving mouse motion.
 .TP
+.BR "-M"
```
hean01-cendio: You should break out local / remote mouse rendering from this pull request into one of its own.

rdesktop.c
```diff
@@ -697,6 +700,16 @@ main(int argc, char *argv[])
 					p++;
 				}
 
+				if (*p == '@')
+				{
+					g_dpi = strtol(p + 1, &p, 10);
```
hean01-cendio: If no DPI specified by argument, could we use DPI fetched from display ?

Would that be an good out of the box experience ?

secure.c
joshudson: Yes. That is the size of the head buffer, and it got larger in the protocol change enabling sending DPI.

doc/rdesktop.1
joshudson: Probably so, but Win10 mouse was completely broken without it. I don't like intermediate steps I can't test.

rdesktop.c
joshudson: It would be, but I cannot write it. My box gets the wrong value for all programs. :(

hean01-cendio: Ok, I found something scary of relevance:

  https://bugs.launchpad.net/ubuntu/+source/xorg-server/+bug/589485

joshudson:
In which case trying to query the value was never going to work for
me. I have multiple monitors at different DPIs; the WM config is
useless.

doc/rdesktop.1
hean01-cendio: So the main issue here is that mouse pointers doesn't work (is not shown?) when connecting to Windows 10 ?

joshudson: Yeah it was. Annoyingly the mouse cursor was only broke about half of the time (I had seen this on 2012R2 and had to double remote to fix it). I think the problem actually got fixed in trunk after I submitted all of this, but the fix won't work right if -m is used.

I couldn't use any of the fixes until they were all in though. The Win10 login screen is affected by the mouse problem.

joshudson:
I went back and checked. The underlying issue causing me to need -M for use local mouse cursor is not fixed. The change will have to be allowed to stand whole.

hean01-cendio:
Could you please cleanup this PR:
 - remove local cursor parts
 - fix the conflicts

jhudsoncedaron:
I'll get to it sooner or later.

joshudson:
The bad news is the new cursor code doesn't seem to work at all on Windows 10. Given my previous success on win10 I suspect that my screen snapshot is from master a few months ago not two weeks ago. The good news is the new cursor code fixed the Windows 7 cursor glitch involving the I-beam.

joshudson:
I'm stumped with the cursor. It doesn't reach process_colour_pointer_common(STREAM s, int bpp) anymore.

joshudson:
Full bisection reveals this falls down if DPI is set to 144 or larger; this is the size at which Windows would replace its 32 bit cursors with 48 bit cursors; and that capability is not declared. CAPSETTYPE_LARGE_POINTER (27) would have to be declared; which in turn requires multifragment update set to at least 38,055. I can't even find the capability code.

The exact behavior of what happens whey trying to exceed the pointer size depends on which version of Windows the client is talking to. This is why this appears to work on Windows 8.1.

I suppose the best I can do is autodetect when this would fail and use local mouse cursor in that case.

kierun:
I just tried that PR. I merged master into it so it's a little more up to date with other things. I tried this:

`./rdesktop -g1920x1080@200 192.168.3.229`

Which got me a window of the right size but no change in DPI.

Did I do something wrong?…

joshudson:
1) changing DPI after login isn't reliable: windows patch level dependent

2) If you accidentally tried it with a binary without the change you won't get an error message.

3) I didn't know you could omit the space after -g so I never tried it.

kierun:
> changing DPI after login isn't reliable: windows patch level dependent

Err… I was not logged in before I changed the DPI. This was to connect to the machine. Or do you mean something else?

> If you accidentally tried it with a binary without the change you won't get an error message.

No, I am sure this is the one in the current directory where I have the repository cloned in. The last commit is `ae31a05f228c7c9213d3d56b2f574c665acf8554`:

```
commit ae31a05f228c7c9213d3d56b2f574c665acf8554                                 
Author: Karl Mikaelsson <derfian@hamsterkollektivet.se>                         
Date:   Fri Oct 6 12:54:11 2017 +0200   

    Let Travis build with both GCC and clang, take 2                            
                    
    Last approach didn't work, hopefully this will work better.
```

> I didn't know you could omit the space after -g so I never tried it.

☺ Makes no difference whatsoever.

joshudson:
Please tell me you weren't silly enough to try this on Windows 7. The protocol extension to do this was added in Windows 8 or 8.1.

kierun:
☺ Nope, Windows 10 Pro VM.

uglym8:
I've just verified these changes on Windows 10 Pro VM.
Everything works as expected except for cursor disappearance on DPIs greater than 200 (see #165 for details).

secure.c
```diff
@@ -440,15 +441,25 @@ sec_out_mcs_data(STREAM s, uint32 selected_protocol)
...
 	out_uint8(s, 0);
 	out_uint32_le(s, 1);
 	out_uint8s(s, 64);
 	out_uint32_le(s, selected_protocol);	/* End of client info */
+	if (g_dpi > 0)
+	{
+		/* Extended client info describing monitor geometry */
+		out_uint32_le(s, g_width * 254 / (g_dpi * 10)); /* desktop physical width */
+		out_uint32_le(s, g_height * 254 / (g_dpi * 10)); /* desktop physical height */
+		out_uint16_le(s, 0); /* orientation: landscape */
```
uglym8: LGTM but IMHO it would be better to add defines for desktop orientations according to https://msdn.microsoft.com/en-us/library/cc240510.aspx i.e. 
#define ORIENTATION_LANDSCAPE 0
#define ORIENTATION_PORTRAIT 90
etc.
What do you think?
P.S. It'd be a good PR policy to add proper defines from corresponding protocol documentation.

secure.c
```diff
@@ -440,15 +441,25 @@ sec_out_mcs_data(STREAM s, uint32 selected_protocol)
...
+		out_uint32_le(s, g_dpi < 96 ? 100 : (g_dpi * 100 + 48) / 96); /* desktop scale factor */
```
uglym8: Maybe add a comment about desktopScaleFactor bounds (100%-500%) too?
Or I just a little bit picky?:)

kierun:
> I've just verified these changes (only DPI related part though) on Windows 10 Pro VM.
> Everything works as expected except for cursor disappearance on DPIs greater than 200 (see #165 for > details).

Hum, so I am doing something wrong. This is my full command line and verbose output:

```
(DPI|…); ./rdesktop -v -u usr@example.com -g 1920x1080@150 192.168.3.229
Password: 
Failed to intialize NLA, do you have correct kerberos tgt initialized ?
Failed to connect using NLA, trying with SSL
Connection successful
Clipboard(error): xclip_handle_SelectionNotify(), unable to find a textual target to satisfy RDP clipboard text request
```

Yet it all appears with tiny fonts… ☹ 

What can I try?

rdesktop.c
```diff
@@ -669,6 +671,16 @@ main(int argc, char *argv[])
...
+					if (g_dpi <= 0)
+					{
+						error("invalid geometry\n");
```
uglym8: Could you change this to logger(Core, Error, whatever you want to show to user)? 
e.g.:
logger(Core, Error, "Wrong DPI value (%d)", g_dpi);

uglym8:
@kierun, can you verify that your source code drop really contains this PR's changes?
What happens when you run rdesktop with negative value for dpi (e.g. -g 1920x1080@-1)?

uglym8:
@kierun, you can checkout my branch (https://github.com/uglym8/rdesktop/tree/check_dpi) containing only DPI-related changes.
Run rdesktop prepending RDESKTOP_DEBUG=Core e.g.:
````bash
RDESKTOP_DEBUG=Core  ./rdesktop -v -u usr@example.com -g 1920x1080@150 192.168.3.229

File: secure.c (hunk 5 above)
Comment (joshudson) [reply secure.c]: I don't believe the upper bound is accurate in the spec.

File: rdesktop.c (hunk 6 above)
Comment (joshudson) [reply rdesktop.c]: Well I could but that check is intended to catch @letters so %d is silly.

Comment (uglym8) [reply rdesktop.c]: Ok.
The point is to use the logger() instead of error().
What you're going to output to user is up to you here actually

**Comment by kierun** [comment]
> You can checkout my branch (https://github.com/uglym8/rdesktop/tree/check_dpi) containing only  DPI-related changes.

Yeah, okay, this works. I wonder what branch I was running. Never mind.

I have no mouse pointer but that seems expected… ☺

**Comment by joshudson** [comment]
@kierun Which is why my patch has mouse code in it. Local pointer is better than no pointer.

**Comment by kierun** [comment]
@joshudson Apologies for being dense … Is your patch in https://github.com/uglym8/rdesktop/tree/check_dpi? If not, where can I get it?

BTW, just in case this was not obvious: *I very much appreciate your and @uglym8's time and effort. Thank you very much indeed.*

**Comment by joshudson** [comment]
@kierun: https://github.com/joshudson/rdesktop branch issue65.

**Comment by kierun** [comment]
@joshudson 

>  https://github.com/joshudson/rdesktop branch issue65.

The first time I ran it, it was fine. The second (with the same parameters), I got this:

```
Failed to intialize NLA, do you have correct kerberos tgt initialized ?
Core(debug): GSS error [0:13:0]: cssp_connect(), negotiation failed.
Core(debug):  - Unspecified GSS failure.  Minor code may provide more information
Failed to connect using NLA, trying with SSL
Connection successful
X Error of failed request:  BadAtom (invalid Atom parameter)
  Major opcode of failed request:  17 (X_GetAtomName)
  Atom id in failed request:  0x0
  Serial number of failed request:  6172
  Current serial number in output stream:  6172
```

I was trying to see if the cursor was not vanishing on subsequent runs…

jhudsoncedaron:
I am pretty sure that error isn't mine or you accidentally tried to use the surge branch. The surge branch has some crazy X corruption on shutdown I haven't tracked down yet.

uglym8:
@joshudson,  @jhudsoncedaron, @hean01-cendio , guys, this PR while is rather straightforward is starting to take a ridiculous amount of time to merge:) 

It works perfectly so only minor (cosmetic) changes are needed before we can merge it.

Let me suggest the following way to resolve this issue: @joshudson, could you please add the requested  (by @hean01-cendio  and me) changes?
Or I can do it by myself by combining my test branch (with your changes) and changes left to fix.

What do you guys think?

joshudson:
@uglym8 So we're actually at an impasse. @hean01-cendio  wants the mouse code out but it won't work without it right now. As more and more cases behave themselves I've been tightening up the cases that disable remote cursors, but I don't think it will ever get to zero.

uglym8:
@joshudson,  first of all, thank you for your contribution.

Regarding us beiing at an impasse: actually quite the opposite:) 
We're making steady progress through this issue alas a little bit slowly.

Back to business: Thank you for your recent changes. LGTM.

One final thought though: I propose to remove mouse pointer related stuff from this PR and continue to discuss it in #165.  
Karl (@derfian) is making impressive progress with the Fast-Update handling issue #157 (See #168).
So I guess we'll soon get this resolved.

Meanwhile those who needs high DPI (over 200) support can use your local cursor changes (that's why I suggest to explicitly mark your changes as temporary workaround in #165).

What do you think?

joshudson:
It turns out you need local cursor anyway if you pass -m. Some cursors (think application specific cursors) are incompatible with other window backgrounds. I'm pretty sure I'm not the only one using -m but I doubt many people use both -m and high contrast black.

hean01-cendio:
Sorry for not commenting earlier on this thread, I have had a fever but is now back in business.

Let me clarify my stand point, I do not have something against local mouse feature and would gladly bring it to into rdesktop as a stand alone feature. Also I do think that the HI DPI feature should no rely on the local cursor feature, that'ss why this PR is not yet merged and we need to get support for large pointers in rdesktop before we merge this branch.

@derfian has been doing a great work of clarifying the implementation of capabilities/fastpath and we are working together to bring multi fragment support as an additional step closer to multi pointer. In fact, we are almost there, and next step is to implement the large pointer support.

With that said, I want @joshudson to break out the 'local cursor' implementation into its own PR and it will merged right away.

derfian:
PR #174 adds support for large pointers so we won't have to fall back on the system default pointers.

large-pointers2

hean01-cendio:
PR #174 is now merged to master and I'm now good with merging this PR if the local mouse pointer is moved into its own pull request.

joshudson:
If anybody here knows how somebody else was able to push to my repository please own up to it before I file a security incident with github. Something was deranged on the server and I had to do git push -f

uglym8:
There's a checkbox named "Allow edits from maintainers." on the right side of the PR window, which do exactly what's written.
So, I reckon it's a false alarm:)

uglym8:
BTW, LGTM now.
Thank you very much,

hean01-cendio:
This works really nice now, just one thing that holds me back for merge. This PR have a long history and bit too many commits. See https://github.com/rdesktop/rdesktop/compare/master...joshudson:issue65.

@joshudson could you do a cleanup of the commits on this PR. You could use an interactive rebase or create a patch of branch against master and reapply than force push it.

Issue #84, opened 2017-01-06
Feature request: Add SO_KEEPALIVE option


rfrht:
Hello there!

I'd like to suggest a switch for enabling SO_KEEPALIVE. Thank you!

hean01-cendio:
This is a server side configuration. See [1] form more details.

[1] https://social.technet.microsoft.com/Forums/windowsserver/en-US/4e86e128-1339-457f-8e37-73c090b95cd5/windows-2008-r2-server-terminal-services-keepalive?forum=winserverTS

rfrht:
Hi Henrik,

Thanks for your consideration.

Of course; The keepalive can be set in the server too. Matter of fact, the keep-alive is configurable on *every* endpoint: Client or server. It is a `socket()` property. If the server does not enables `SO_KEEPALIVE`, the server will honour client-initiated keep-alive packets.

And the reverse is also true: If the server has `SO_KEEPALIVE` enabled but the client is not, the client will correctly reply the server's keepalive replies.

Please allow me add some more evidence so you can reconsider it. The following evidence was taken with a client compiled using the patch.

~~~
[rfreire@rf ~]$ rdesktop 192.168.122.83 &
[rfreire@rf ~]$ netstat -nto | grep 3389
tcp        0      0 192.168.122.1:39806         192.168.122.83:3389         ESTABLISHED off (0.00/0/0)
~~~

As you may see above, the client is not originating any keepalive; it will be completely dependent by Server's keepalive configuration.

Now see the below example:

~~~
[rfreire@rf ~]$ rdesktop 192.168.122.83 &
[rfreire@rf ~]$ netstat -nto | grep 3389
tcp        0      0 192.168.122.1:39818         192.168.122.83:3389         ESTABLISHED keepalive (2.41/0/0)
~~~

Now we have the keepalive enabled and the server will reply the **client's** keepalive settings.

What am I trying to fix here:

During the holidays season I was working using a different ISP, and in this ISP I was behind a [CGNAT](https://en.wikipedia.org/wiki/Carrier-grade_NAT).
The carrier NAT options were pretty aggressive; if the connection were idle for 5 minutes, the NAT entry was dropped at the carrier, and the connection lost to my remote Windows session.

Then, I rebuilt `rdesktop` adding keepalive settings and set some aggressive values, such as: 

~~~
net.ipv4.tcp_keepalive_time=120
net.ipv4.tcp_keepalive_probes=5
[...]

PR #85, opened 2017-01-06
rdesktop: Add TCP Keepalive switch

rfrht:
Resolves Issue #84 

Added switch -J to enable TCP Keepalive in the client side, so it can
take advantage of TCP keeping session when traversing aggressive
firewalls.
By default it is disabled (current behaviour).
For more information on TCP Keepalive, please visit
http://tldp.org/HOWTO/TCP-Keepalive-HOWTO/overview.html

Signed-off-by: Rodrigo Freire <root@rf01.pro>

### DIFF
--- a/rdesktop.c
+++ b/rdesktop.c
@@ -574,7 +574,7 @@ main(int argc, char *argv[])
 #define VNCOPT
 #endif
 	while ((c = getopt(argc, argv,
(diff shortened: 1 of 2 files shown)


hean01-cendio:
This is a server side configuration. See [1] form more details.

[1] https://social.technet.microsoft.com/Forums/windowsserver/en-US/4e86e128-1339-457f-8e37-73c090b95cd5/windows-2008-r2-server-terminal-services-keepalive?forum=winserverTS

PR #116, opened 2017-04-28
Allow password being provided via env variable RDESKTOP_PASSWORD

m4tze:
For some reasons I need to provide the password via process environment. This is

a) more secure than providing it via cmd (i know that you X the password later)
b) very convenient.

### DIFF
--- a/rdesktop.c
+++ b/rdesktop.c
@@ -60,6 +60,7 @@ uint8 g_static_rdesktop_salt_16[16] = {
 
 char g_title[64] = "";
 char *g_username;
+char *env_password;
 char g_password[64] = "";
 char g_hostname[16] = "";
 char g_keymapname[PATH_MAX] = "";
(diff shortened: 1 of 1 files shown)


jhudsoncedaron:
a) more secure than providing it via cmd 

Not really. ps can get environment variables. Consider echo "passord" | rdesktop -p - which is secure if echo is a shell builtin.

m4tze:
ps gets the environment via procfs - therefore it is not possible to get the env of another user (being root is another story). This is, in contrast to the commmand line, a huge security advantage.

hean01-cendio:
@m4tze, my standpoint on this PR is that I want to close it in favor of enhancements in issue #127. With askpass you could have a wrapper that reads password from environment if you are bound to do that.

In meantime you could just do the following to fullfill you requirement of using password from environment:

```
echo $RDESKTOP_PASSWORD | rdesktop -p - <servername>
```

PR #131, opened 2017-07-10
Use standard GSSAPI rather than libgssglue

CendioOssman:


opoplawski:
Any chance this could get applied to the v1.8 branch?  The Fedora rdesktop package is lacking CredSSP support because of it.

Issue #128, opened 2017-07-07
Migrate from OpenSSL to GnuTLS, Nettle and libtasn1


hean01-cendio:
The licensing of OpenSSL is questionable and it is an all in one solution for crypto. It would be very nice to migrate away to components such as GnuTLS, nettle and libtasn1 for a cleaner and more isolated code.

Any thought about this ?

uglym8:
AFAIU, libreSSL has the same dual licence as OpenSSL does (I was gonna suggest it for using).
I was not able to find any *terrible* news regarding GnuTLS so maybe it's really a good choice and worth trying.

hean01-cendio:
GnuTLS branch is now merged into master

Issue #13, opened 2015-10-25
-f option issue


zhanghqGit:
rdesktop -f ip ,full screen mode.
when connect is down by network broken,what can i do is just force shutdown my computer.

hean01-cendio:
See bug #17 for information

Issue #11, opened 2015-10-01
There is no ./configure script


douglasheld:
I would like to build on Apple OS X and there is no configure script as documented in the README.

hean01-cendio:
The information how to build snapshot of the source is available in the README. However it was not reflecting the move to GitHub.

README is updated in commit d529a86

douglasheld:
Thank you!
Success steps:
sudo yum install autoconf libgssglue-devel automake libX11-devel openssl-devel
./bootstrap
./configure --disable-smartcard
make
sudo make install

Issue #10, opened 2015-09-15
Mouse scroll wheel not working


antoinetran:
Hi,

I am using Red Hat 6.5 with rdesktop 1.8.2 (tested with 1.8.3 too) and xrdp 0.5.0-0.13.el6 (tested with 1.6.1-3.el6 too). I tried to connect to the server using rdesktop but the scroll wheel does not work. xev does not show any reactions when scrolling up or down. But using middle button works. What do I have to do for the server to receive wheel events?

Here are some useful details:
a) Windows native rdesktop client -> linux VM xrdp 0.5.0 scroll OK
a) Linux rdesktop 1.8.2 -> Windows VM scroll OK

Thank you.

hean01-cendio:
I took a look into this issue. RDP have two ways of sending keybord/mouse events named as slow-path and fast-path. As I understand fast-path was added to the protocol to take away some overhead in these input event packets. Both ways are supported.

rdesktop only imlpements slow-path and therefor I think that you found a bug in xrdp slow-path handling due to that rdesktop works good with a window terminal server using slow-path sending of key/mouse events.

antoinetran:
Hi,

Thank you for your answer. Is there any workaround and/or any logs I can find to help?

hean01-cendio:
From what I can tell the problem lies within the slow path input handling implementation in xrdp.

PR #137, opened 2017-08-08
Replace OpenSSL with GnuTLS, Nettle, libtasn1 and GMP

uglym8:
See #128


hean01-cendio:
Nice work!, I will check this PR within this week for merge as soon as possible.

asn.c
```diff
@@ -17,8 +18,20 @@
...
 
+#include <gnutls/gnutls.h>
+#include <libtasn1.h>
+#include <stdlib.h>
+
+#include "asn.h"
+
 #include "rdesktop.h"
+// Generated by asn1Parser
+#include "pkix_asn1_tab.c"
```
CendioOssman: This file isn't included in this commit, so it will create a broken commit. Please merge this commit and the next one.

Also, we should probably have a configure test for libtasn1?

configure.ac
```diff
@@ -29,6 +29,11 @@ fi
     
 AC_PATH_TOOL(PKG_CONFIG, pkg-config)
 
+# no .pc for GMP
+AC_SEARCH_LIBS([__gmpz_init], [gmp])
+# pkg-config file Nettle doesn't provide hogweed in libs
+AC_SEARCH_LIBS([nettle_rsa_public_key_clear], [hogweed])
```
CendioOssman: These seem like hacks for static linking rather than something needed for a "normal" link?

configure.ac
```diff
@@ -160,6 +96,48 @@ if test x"$HAVE_XRANDR" = "x1"; then
     AC_DEFINE(HAVE_XRANDR)
 fi
 
+# libtasn1
```
CendioOssman: Can we split these tests up and include them with the commit that actually starts using code from each library?

ssl.h
```diff
@@ -22,25 +23,26 @@
...
 
-#include <openssl/rc4.h>
-#include <openssl/md5.h>
-#include <openssl/sha.h>
-#include <openssl/err.h>
-#include <openssl/bn.h>
-#include <openssl/x509v3.h>
-#include <openssl/hmac.h>
+#include <gnutls/x509.h>
+#include <gnutls/abstract.h>
```
CendioOssman: These look like they belong to a different commit?

ssl.h
```diff
@@ -22,25 +23,26 @@
...
-#define D2I_X509_CONST
-#endif
+#include <nettle/md5.h>
+#include <nettle/sha1.h>
+#include <nettle/arcfour.h>
+#include <nettle/hmac.h>
+#include <nettle/rsa.h>
+
+#include <gmp.h>
+#include <libtasn1.h>
```
CendioOssman: Do we really need all of these headers here? It's good to keep the dependency chains as small as possible.

tcp.c
```diff
@@ -303,122 +281,149 @@ RD_BOOL
...
-	}
 
-	if (SSL_set_fd(g_ssl, g_sock) < 1)
-	{
-		logger(Core, Error, "tcp_tls_connect(), SSL_set_fd() failed");
-		goto fail;
+	} else {
+		char *desc;
+		desc = gnutls_session_get_desc(g_tls_session);
+		//printf("TLS  Session info: %s\n", desc);
```
CendioOssman: Please don't include dead code.

tcp.c
```diff
@@ -537,14 +542,13 @@ tcp_connect(char *server)
...
-	if (g_ssl)
-	{
-		if (!g_network_error)
-			(void) SSL_shutdown(g_ssl);
-		SSL_free(g_ssl);
-		g_ssl = NULL;
-		SSL_CTX_free(g_ssl_ctx);
-		g_ssl_ctx = NULL;
+	if (g_ssl_initialized) {
+		//gnutls_bye(g_tls_session, GNUTLS_SHUT_RDWR);
```
CendioOssman: Why is this commented out?

ssl.c
```diff
@@ -92,142 +92,172 @@ void
...
+ * Upon detecting "problem" (wrong public RSA key OID) certificate
+ * we basically have two options:
+ *
+ * 1)) encode certificate back to DER, then parse it using libtasn1,
+ * fix public key OID (set it to 1.2.840.113549.1.1.1), encode to DER again
+ * and finally reparse using GnuTLS
+ *
+ * 2) encode cert back to DER, get RSA public key parameters using libtasn1
+ *
+ * Or can rewrite the whole certificate related stuff later.
```
CendioOssman: Any tweak flags or something to GnuTLS that can make it compensate for this? Their developer is usually very responsive so it could be worth asking on their mailing list.

ssl.c
```diff
@@ -92,142 +92,172 @@ void
...
+		}
 
-		return NULL;
-	}
+		/* By some reason, Microsoft sets the OID of the Public RSA key to
+		   the oid for "MD5 with RSA Encryption" instead of "RSA Encryption"
 
-	nid = OBJ_obj2nid(algor->algorithm);
+		   Kudos to Richard Levitte for the finding this and proposed the fix
+		   using OpenSSL. */
```
CendioOssman: I guess the last part is no longer relevant?

.travis.yml
```diff
@@ -1,5 +1,11 @@
+dist: trusty
 language: c
 
 compiler: gcc
 
+before_install:
+ - sudo apt-get -qq update
+ - sudo apt-get install -y nettle-dev libtasn1-dev libgnutls-dev
```
CendioOssman: Please split this out to all the relevant commits.

asn.c
uglym8: Sure.
I did not consider to make each commit distinct and buildable as the resulting commits are the product of merging and rebase anyway of commits in various privates branches.
So I think I can reorder and merge commits as you requested.

configure.ac
uglym8: Yep. Got the same feeling but it's the best I could come up with as pkg-config for nettle doesn't return -lhogweed (for --libs). My google-fu didn't help either:)
So if this is a real showstopper I could devote some time to try something more decent.

configure.ac
uglym8: Sure. Same answer as for asn.c

ssl.h
uglym8: Will check thoroughly and remove if we could live without these includes.

tcp.c
uglym8: Because when connecting to some of my test windows machines the connection hangs after calling this one and I had cheated by commenting it out as I had no powers left to debug this issue.
As  I have said above it this a showstopper I can debug and fix this issue.

ssl.c
uglym8: AFAIK, there aren't any.

The last discussion I saw regarding this matter is  of  2012:
https://lists.gnupg.org/pipermail/gnutls-help/2012-August/002837.html

Yes, I contemplated about the possibility to fix this (as it's rather trivial) and send the PR to the GnuTLS folks but I don't think that this fix will land in major distros soon.

So the end result would be that I have to fix this issue in upstream GnuTLS and in rdesktop (covering the cases when distro doesn't have the recent enough GnuTLS version).

So it thought the current approach is more pragmatic.

ssl.c
uglym8: I've left it as token of gratitude to the person who first found it and fixed.
Sure it could be removed but I'd prefer to leave it.
Anyway, I'm open to suggestions:)

.travis.yml
uglym8: Sure.

uglym8:
So basically I should make all requested changes and send another PR based on the current HEAD?
Right?

tcp.c
uglym8: I left it intentionally to show future me or anyone reading this code how to get session info.
Sure It could be ditched.

hean01-cendio:
@uglym8 yes, that sounds correct, now when its about which commits includes what, rather small changes.

configure.ac
CendioOssman: I'd say leave it out. People are generally expected to provide their own link information when they are doing static linking.

tcp.c
CendioOssman: Please have another look. It seems very wrong to not do a proper shutdown. If you get stuck then at least include a FIXME comment that this should be investigated further.

ssl.c
CendioOssman: Alright, let's go with this for now

ssl.c
CendioOssman: Alright, but an amendment might be in place that we've since switched to GnuTLS.

CendioOssman:
You can do a force push to your existing branch and GitHub should sort things out. You should not have to open another PR.

configure.ac
uglym8: These checks aren't  for static linking. They are primarily used for dynamic linking.
Check output of the following command:
```bash
 pkg-config --libs nettle
````

My distro shows just one dep: 
````bash
-lnettle
````

Which is obviously not enough for proper linking for program using Nettle (lacks -lhogweed).
Same applies to GMP case.

Or I'm missing something here?

File: tcp.c (hunk 7 above)
Comment (uglym8) [reply tcp.c]: SUre. I'll do it because I don't like this approach either.

File: configure.ac (hunk 2 above)
Comment (CendioOssman) [reply configure.ac]: What errors are you getting?

Comment (uglym8) [reply configure.ac]: Try removing Nettle related string and execute:
````
 ./bootstrap && ./configure && make
````

I'm getting:
````
/usr/bin/ld: ssl.o: undefined reference to symbol 'nettle_rsa_public_key_clear@@HOGWEED_4'
/usr/lib/libhogweed.so.4: error adding symbols: DSO missing from command line
````

Comment (CendioOssman) [reply configure.ac]: Alright, the problem is that we are using hogweed methods and not just nettle methods. They are somewhat independent from what I can see in the documentation.

hogweed has a pkg-config file, so please update the configure script to also include that.

Comment (uglym8) [reply configure.ac]: Ok. You're right. I'll fix this one.
What about GMP check? Can I leave it as it is now?

Comment (CendioOssman) [reply configure.ac]: Not really. Do we call any gmp functions directly? If so we can have the check, but we should check for a more sensible function rather than something internal like `__gmpz_init`.

Comment (uglym8) [reply configure.ac]: Check rdssl_rsa_encrypt() in ssl.c for mpz_* functions.
Even it starts with __ It's not internal function btw:
````
$ readelf -a /usr/lib/libgmp.so|grep mpz_init$
    92: 000000000001f280    37 FUNC    GLOBAL DEFAULT   12 __gmpz_init
`````
Bad name, sure.

File: tcp.c (hunk 7 above)
Comment (uglym8) [reply tcp.c]: Basically there was an unfortunate combination of two bugs:
1) I'd better use ( it's way faster)
````C
gnutls_bye(g_tls_session, GNUTLS_SHUT_WR)
````
 instead of
````C
gnutls_bye(g_tls_session, GNUTLS_SHUT_RDWR)
````
2) I zeroed rcvd in **tcp_recv**() forming a perfect infinite loop:)
Check **tcp.c: 240-245**

**Comment by uglym8** [comment]
Fixed.
Please review again.

File: tcp.c [hunk 11]
```diff
@@ -537,14 +555,15 @@ tcp_connect(char *server)
...
-	{
-		if (!g_network_error)
-			(void) SSL_shutdown(g_ssl);
-		SSL_free(g_ssl);
-		g_ssl = NULL;
-		SSL_CTX_free(g_ssl_ctx);
-		g_ssl_ctx = NULL;
+	int rv;
+	if (g_ssl_initialized) {
+		//rv = gnutls_bye(g_tls_session, GNUTLS_SHUT_RDWR);
```
CendioOssman: I guess this should have been removed now? :)

uglym8: Oh my God:)))))
Yes, it definitely should go now.
Will fix ASAP;)

uglym8:
Fixed and force pushed again.
Sorry, completely forgot about this comment:)

kekePower:
This patch needs to be rebased.

Patch #0 (rdesktop-PR137.patch):
+ /usr/bin/patch --no-backup-if-mismatch -p1 -b --suffix .pr137 --fuzz=0
patching file .travis.yml
Hunk #1 FAILED at 7.
1 out of 1 hunk FAILED -- saving rejects to file .travis.yml.rej
patching file asn.c
patching file asn.h
patching file configure.ac
Hunk #1 FAILED at 153.
1 out of 1 hunk FAILED -- saving rejects to file configure.ac.rej
patching file pkix.asn
patching file pkix_asn1_tab.c
patching file .travis.yml
Hunk #1 FAILED at 7.
1 out of 1 hunk FAILED -- saving rejects to file .travis.yml.rej
patching file configure.ac
Hunk #1 FAILED at 167.
1 out of 1 hunk FAILED -- saving rejects to file configure.ac.rej
patching file ssl.c
Hunk #3 succeeded at 312 (offset 9 lines).
patching file ssl.h
patching file .travis.yml
Hunk #1 FAILED at 7.
1 out of 1 hunk FAILED -- saving rejects to file .travis.yml.rej
patching file configure.ac
Hunk #1 FAILED at 181.
1 out of 1 hunk FAILED -- saving rejects to file configure.ac.rej
patching file tcp.c
Hunk #4 succeeded at 74 (offset 1 line).
Hunk #5 succeeded at 120 (offset -6 lines).
Hunk #6 succeeded at 129 (offset -6 lines).
Hunk #7 succeeded at 182 (offset -6 lines).
Hunk #8 FAILED at 221.
Hunk #9 FAILED at 231.
Hunk #10 succeeded at 269 (offset -6 lines).
Hunk #11 succeeded at 293 (offset -6 lines).
Hunk #12 succeeded at 369 (offset -6 lines).
Hunk #13 succeeded at 566 (offset -6 lines).
2 out of 13 hunks FAILED -- saving rejects to file tcp.c.rej
patching file configure.ac
Hunk #2 FAILED at 54.
Hunk #3 FAILED at 184.
2 out of 3 hunks FAILED -- saving rejects to file configure.ac.rej
patching file ssl.c
Hunk #2 FAILED at 92.
Hunk #3 succeeded at 254 (offset 3 lines).
1 out of 3 hunks FAILED -- saving rejects to file ssl.c.rej
patching file ssl.h
patching file asn.c
patching file asn.h

uglym8:
@kekePower,  yes it should be rebased. 
Got it rebased couple weeks ago and was planning to resubmit it on this weekend.

kekePower:
@uglym8
Thanks for getting back to me. Will wait for your rebased version.

I was going to give this PR a try to add to the tester base.
Was also going to ask my packaging mentor for Mageia if he wanted to push this update to Cauldron, but will now wait until the patch is rebased.

uglym8:
@kekePower, rebased upon the latest master so you can test it.

Old branch is still here as **old_replace_openssl**
Don't be alarmed of change commit count (was 5, now  - 4) as I've merged two last commits.

kekePower:
@uglym8 
Thanks. Patch applied cleanly and rdesktop compiled without errors. Now it's time to test.

kekePower:
@uglym8 
I just installed the patched version and was unable to successfully log on to Windows in a vmware session. Then tried an unpatched version and it worked.

What do I have to do to get the patched version to work? Are there any settings, files or something that has to be configured, either in Linux or in Windows?

$ rdesktop -vvv -a16 -kno virtuality
Password:
Failed to connect using NLA, trying with SSL
Failed to connect using SSL, trying with plain RDP.
Connection established using plain RDP.
Protocol(warning): rdp_loop(), unhandled PDU type 14 received
Core(error): tcp_recv(), recv() failed: Connection reset by peer

Thanks.

uglym8:
Hmm, I've quick tested on ArchLinux after rebase with Win7 and Win2K3.
Seemed to work just fine.
I'll test tomorrow with W10.

What Linux distro are you testing this build on?
What version of the distro?

uglym8:
Ah, got an error with Win7 Ultimate.
Will investigate and fix:)

kekePower:
@uglym8 
I'm using Mageia 6 x86_64, trying to connect to Win7 in vmware 14.

Will try another compile using libao as the sound server instead of libpulse as I had previously.

kekePower:
Another funny message at the bottom...

$ rdesktop -5 -v -a16 -kno virtuality
Password:
Failed to connect using NLA, trying with SSL
Failed to connect using SSL, trying with plain RDP.
Connection established using plain RDP.
Core(error): tcp_send(), send() failed: Connection reset by peer
Connection successful
Disconnecting...
Disconnected due to network error, retrying to reconnect for 70 minutes.
Failed to connect using NLA, trying with SSL
Failed to connect using SSL, trying with plain RDP.
Connection established using plain RDP.
Protocol(warning): rdp_loop(), unhandled PDU type 2 received
Core(error): tcp_recv(), recv() failed: Connection reset by peer
Tried to reconnect for 70 minutes, giving up.

uglym8:
Ok. Thank you for your help.
I'm gonna fix this tomorrow and  report as soon as it's ready.

uglym8:
It's a certificate error as reported by GnuTLS: "Key usage violation in certificate has been detected".

kekePower:
Is this something you can fix in rdesktop or something I have to do in Win7?

hean01-cendio:
> Is this something you can fix in rdesktop or something I have to do in Win7?

@kekePower This means that the certificate (server side, eg. Windows 7) used for RDP is invalid to be used for TLS data encryption. If you haven't set this certificate by your self or via AD, I think Windows generates a self signed default certificate to be used and in that case it is broken. Try to update the ceritficate with a proper one, see [1] for details.

[1] https://support.microsoft.com/en-us/help/2001849/how-to-force-remote-desktop-services-on-windows-7-to-use-a-custom-serv

uglym8:
@kekePower, yes, @hean01-cendio is right. GnuTLS is way more restrictive (or should we say, OpenSSL is more relaxed regarding certificate validation).

Anyway, GnuTLS FAQ  states (https://gnutls.org/faq.html#key-usage-violation):
"Because this misconfiguration problem is widespread, other TLS/SSL implementations used by popular browsers tolerate the violation, and several servers negotiate ciphersuites not allowed by the certificate, newer versions of GnuTLS will also allow such key usage violations (and will only output a warning message). "

So I'm gonna to investigate this a little bit further.

kekePower:
So this patch, as it stands now, basically breaks rdesktop for a lot of Windows 7 users.
This means it can't be distributed until @uglym8 has come up with a less restrictive way of doing authentication.

Or, good documentation regarding this issue has to be created with detailed instructions on how to fix this.

@hean01-cendio Thanks for the documentation. I'll give it a good read and see if I can get it to work.

uglym8:
There's a good discussion about the this matter:
https://lists.gnupg.org/pipermail/gnutls-help/2005-March/000210.html

uglym8:
Actually it's quite easy to fix.

I've changed the priority to **"NORMAL:%COMPAT"** (https://gnutls.org/manual/html_node/Priority-Strings.html).

But I prefer to use it only in a case when handshake with default priority fails:
we should show the warning about "wrong" certificate and try to handshake again with "COMPAT" priority.

I can't do it solely in the **tcp_tls_connect()** as we lose connection in a case of fatal error during handshake.

Long story short: ATM (for testing purpose only) I use "NORMAL:%COMPAT" as a default priority.

Please, pay attention, as it's TEMPORARY workaround.

uglym8:
@kekePower, you can try to test again, btw:)

hean01:
@uglym8 , I have verified that COMPAT solves the issue with invalid key usage. Tested against both Windows 7 with auto generated selfsigned certificate and Windows 2016 using a CA generated RDP certificate.

hean01:
> I can't do it solely in the tcp_tls_connect() as we lose connection in a case of fatal error
> during handshake.

I can't see why this can't be done, and I'm thinking:

Try first with restrictive priority string, if handshake fails, set COMPAT, log a warning about the bad fallback and redo handshake. 

@uglym8 , Could you explain why this can't be done all in tcp_tls_connect() ?

uglym8:
> > I can't do it solely in the tcp_tls_connect() as we lose connection in a case of fatal error during handshake. 

> I can't see why this can't be done, and I'm thinking: Try first with restrictive priority string, if handshake fails, set  COMPAT, log a warning about the bad fallback and redo handshake.

> @uglym8 , Could you explain why this can't be done all in tcp_tls_connect() ?

Sure.
If the handshake fails we lose TCP connection to the host. 
Moreover there's no way (at least I can't find any yet) to conduct another handshake without reconnecting first.

So there's no way I can reestablish the connection inside **tcp_tls_connect()** as the connection is established in other upper-level function(s).

The current approach I'm investigating is to add one more parameter (**use_compat**) to the **tcp_tls_connect()** and **cspp_connect()** and add distinct error value (e.g. **-2** in a case when handshake fails with "strict" (default: **NORMAL**)  priority) to indicate that we should retry handshake with more relaxed priority. ATM I can successfully pass the handshake phase (on the second try i.e. with "**NORMAL:%COMPAT**") but got some issues later on. So the work is in progress for now:)

Sure, I can set some global var and wait for connection to succeed (in the **rdp_loop**()) but again it seems to me that it is not the right way to do it as we change the mechanism (**SSL, HYBRID** etc) on each iteration.

uglym8:
Additional info regarding rehandshake:

I could try to rehandshake in **tcp_tls_connect()** but it may fail as server may legitimately refuse rehandshake and close the connection.

Actually this is what I'm observing ATM.

The only possible way to deal with  this situation is to reconnect to the server and try to handshake again.
But I can't do it solely in **tcp_tls_connect()** (i.e. using **tcp_disconnect/tcp_connect** pair) as I have to connect to all layers (mcs, iso etc - check **rdp_connect()**).

So IMHO the current options are:
1) Use **COMPAT** as default choice unconditionally
2) Use **COMPAT** as default choice and provide additional CLI option (e.g. **strict-tls-checks**) to override the default priority.

**Note:** Also we can try to rehandshake here (i.e. do our best) and bail if server drops connection.

3) Use default priority and try to rehandshake.
Ff the server refuses rehandshake just error out with message to user 
"use **-tls-compat-mode** option" to overcome this issue i.e. other way round than (2).

4) Universal answer: rewrite everything;) 
e.g. add **use_compat** and do what I've already described in my previous messages.

uglym8:
Rebased onto the current master.

Linux build is OK.
OSX build is failing (have to find out the correct names for required packages).

hean01-cendio:
I have resolved the conflicts and added a new PR for moving from OpenSSL to GnuTLS.

I will close this PR and we continue on the newly added one.

PR #145, opened 2017-09-10
Increase maximum password length, and other fixes.

### LINKED ISSUE #144: Increase maximum password length
Because certain three-letter agencies have demonstrated an ability to crack 32-character passwords in real-time, my employer has mandated that all passwords be a minimum of 64 randomly-generated characters.  When I set a 64-character password for Windows, I could no longer log in via rdesktop.  Some investigation revealed a hard-coded 63-byte password length limitation, as the character array is declared with length 64, including the null-byte terminator.

Windows supports passwords of up to 256 bytes (128 wide-byte characters), so rdesktop ought to do the same.

It would be better yet to support 256 bytes *by default*, and allow the limit to be increased or decreased via compilation switches.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/49178263-increase-maximum-password-length?utm_campaign=plugin&utm_content=tracker%2F19353203&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F19353203&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pillarsdotnet:
* Resolves #144 by setting maximum host, password, and file path length via autoconf defines.
* Fixes a minor spelling error in the manpage.
* Uses [fputs](https://linux.die.net/man/3/fputs) instead of [fprintf](https://linux.die.net/man/3/fprintf) where there is no format string.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -531,6 +531,18 @@ size_t iconv();
 UTILS_FUNC_ICONV
 LIBS="$LIBS $LIBICONV"
 
+dnl Maximum hostname length.
+AC_MSG_CHECKING([HOST_NAME_MAX])
+AC_CHECK_DECL([HOST_NAME_MAX], , [AC_DEFINE([HOST_NAME_MAX], [64])], [include <unistd.h>])
+
+dnl Maximum password length.
+AC_MSG_CHECKING([PASS_MAX])
+AC_CHECK_DECL([PASS_MAX], , [AC_DEFINE([PASS_MAX], [256])])
+
+dnl Maximum path length.
+AC_MSG_CHECKING([PATH_MAX])
+AC_CHECK_DECL([PATH_MAX], , [AC_DEFINE([PATH_MAX], [256])])
+
 #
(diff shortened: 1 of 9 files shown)


pillarsdotnet:
@hean01-cendio -- Can I get a review of this?

hean01-cendio:
I have already reviewed the code. One thing only, did you verify that the protocol were the strings are used is dynamic lengthed strings ?

CendioOssman:
It seems like overkill (or even possible foot shooting) to have these configurable. Can't we just adjust these to the protocol/system limitations and be done with it? `HOST_NAME_MAX` and `PATH_MAX` should both come from `limits.h` and shouldn't require special handling at all.

hean01-cendio:
I'm closing this pull request due to it's the wrong take on the problem, also it includes unrelated changes. Added more information on issue #144.

Issue #86, opened 2017-01-12
Mouse Cursor Invisible


bendparker:
I am running RPi 3, and rdesktop 1.8.2, connecting to Windows Server 2012 R2.  Upon connection the mouse pointer becomes invisible.  The "disable pointer shadow" does not work.  I've tried to manually set the pointer to something else and that doesn't work either.  It's either always invisible, or it's stuck on the black line text selection cursor.

I apologize if this has been fixed, but I've tried all the fixes listed and nothing has worked.

hean01-cendio:
Do you know which bitdepth the X server is using ?

Could you try run rdesktop with '-a 32' argument to see if that matters ?

bendparker:
Yes I will try this.

bendparker:
Some more information:  I changed from full screen mode to a slightly less than full screen (-g 1900 x 1040) and this helped a bit.  The mouse pointer stayed, but now seems to be intermittently disappearing.  Trying the -a 32 and will report back.

bendparker:
I tried the suggested fix and it did not work.  A little more info, the mouse cursor seems to disappear when typing an email (or any other larger text box) and then never returns.

jhudsoncedaron:
Sounds like the same bug that caused me to add -M on my branch to not accept remote mouse cursors.

bendparker:
Does it work for you?  If so I can try your branch?  Or can I just add -M to my script on the main branch?

bendparker:
I have actually relegated this to a particular user's profile.  If I do the exact same connection with another user, it works fine.

jhudsoncedaron:
-M does not exist on the main branch. My branch is here: https://github.com/joshudson/rdesktop

hean01-cendio:
Could it bew related to mouse cursor themes that includes alpha which rdesktop can't render ? If it works with another one user profile but not a second ?

bendparker:
I think it has to do with some weird graphic settings that he uses.  He likes the text and such to appear larger than they really are.

hean01-cendio:
Nice hints, then we have something to go on. Is he using high dpi mode eg. scaling of everything within his session ?

bendparker:
Yes.  The "Make Text appear larger" option within view properties.

jhudsoncedaron:
That would also explain why I can't not encounter it. I must use high DPI.

dad68:
@bendparker 
(sorry for my English....)
[...]

PR #147, opened 2017-09-15
Fix order of colour values in BGR and BGRA cursor types

imc0:
rdesktop has some problems rendering the cursors from xorgxrdp and some Windows systems.  One problem is this: the colour data is being interpreted the wrong way round.  The order of components for each pixel in a 32-bit cursor image is: B, G, R, A (this translates into ARGB when loaded as a little-endian 32-bit integer).  Currently rdesktop is getting this wrong by loading A as one of the colour values, so it merges the alpha channel with the G and R channels to produce the cursor.

### DIFF
--- a/xwin.c
+++ b/xwin.c
@@ -2944,12 +2944,12 @@ get_next_xor_pixel(uint8 * xormask, int bpp, int *k)
 			break;
 		case 24:
 			s8 = xormask + *k;
-			rv = (s8[0] << 16) | (s8[1] << 8) | s8[2];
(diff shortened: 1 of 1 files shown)


hean01:
Interesting findings, I wonder if this is related to issue #86. I will look into this the upcoming week.

hean01-cendio:
When I looked into this I got really confused and dug down to the main problem. Now the cursor code is rewritten in and now correctly supports rgb and alpha cursors. See commit 6dd85edc6f82.

PR #148, opened 2017-09-18
Obey the Alpha instead of the AND-mask for 32bpp cursors

imc0:
(Following on from PR # 147)
rdesktop has some problems rendering the cursors from xorgxrdp and some Windows systems.  I am not sure about Windows, but I know that xorgxrdp does not supply a useful AND mask when it transmits a cursor with an Alpha channel - in fact it leaves the AND-mask as all zeros, which makes rdesktop display the cursor in a black box.  I do not know what the spec requires (or if there even is a spec) but other RDP clients are ignoring the useless AND mask and using the Alpha channel instead.  Here is one possible way of doing this: it treats the AND mask as opaque if and only if the opacity value is above some threshold (I've chosen 0xf0 which seems to work with xorgxrdp but this is somewhat arbitrary).  Ultimately it would be good to get rdesktop to support colour cursors as FreeRDP does, but I don't think I have development time for that.

### DIFF
--- a/xwin.c
+++ b/xwin.c
@@ -2949,7 +2949,7 @@ get_next_xor_pixel(uint8 * xormask, int bpp, int *k)
 			break;
 		case 32:
 			s8 = xormask + *k;
-			rv = (s8[2] << 16) | (s8[1] << 8) | s8[0];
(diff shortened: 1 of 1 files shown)


imc0:
(Apologies: I've accidentally included the commit for PR#147 here as well)

PR #156, opened 2017-10-04
Don't log common Smart Card error codes as rdesktop errors

derfian:
The proper thing to do would be to forward the errors to the server
and let it deal with it. No need to present these as errors to the
user.


derfian:
I'm considering whether we need a failure log from TS_ScardGetStatusChange at all.

PR #162, opened 2017-10-05
Fix osx build

uglym8:
Fix building rdesktop on macOS X:
- Use Kerberos framework for CredSSP
- No SCARD_CTL_CODE(code) in Apple's PC/SC
- No HOST_NAME_MAX on OS X

To verify:
1) Install Homebrew
2) brew install autoconf automake pkg-config openssl
3) Install XQuartz (X11)
4) build with either Apple supplied PC/SC or libpcsclite (install pcsclit using brew and set PCSCLITE_LIBS and PCSCLITE_CFLAGS to point to install location)

````bash
$ ./bootstrap
$ PKG_CONFIG_PATH=/opt/X11/lib/pkgconfig ./configure --with-openssl=/usr/local/opt/openssl 
$ make

I built and verified that I can connect to Windows 7 box with either one or two smartcard readers.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -123,9 +123,19 @@ fi
 dnl CredSSP feature
 AC_ARG_ENABLE([credssp], AS_HELP_STRING([--disable-credssp], [disable support for CredSSP]))
 AS_IF([test "x$enable_credssp" != "xno"], [
-	  if test -n "$PKG_CONFIG"; then
-	    PKG_CHECK_MODULES(GSSAPI, krb5-gssapi, [WITH_CREDSSP=1], [WITH_CREDSSP=0])
-	  fi
+	case "$OSTYPE" in
+	     darwin*)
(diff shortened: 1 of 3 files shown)


hean01-cendio:
Built rdesktop usigng provided info and verified that I could authenticate against Windows 2016 using smart card on macOs 10.12.6.

hean01-cendio:
Could not test kerberos authentication due to following error:

    lab-136:rdesktop cendio$ RDEKSTOP_DEBUG=Smartcard ./rdesktop -d lab -r scard  rds11.lab.lkpg.cendio.se
    Autoselecting keyboard map 'sv' from locale
    Password: 
    Core(error): cssp_connect(), confidiality service required but is not available
    Redirected to cendio@10.48.3.6 session 17.
    Failed to intialize NLA, do you have correct kerberos tgt initialized ?

Issue #173, opened 2017-10-16
Support large pointers


derfian:
RDP supports up to 96x96 px cursors, but the client needs to advertise support for this. It also requires support for fragmented fast-path updates.

PR #174, opened 2017-10-16
Enable support for large pointers

derfian:
The cursor-handling code already handles larger pointer sizes. This
advertises that rdesktop has the capability to handle large cursors
and adjusts the maximum size of fragmented packets to suit the large
cursor requirements.

Solves issue #173.


hean01:
LGTM!

Looks exactly like the code we sat with except that we missed to increase number of caps :)

PR #176, opened 2017-10-18
Clarify MCS Connect Initial packet

derfian:
Rename the function sending it, use constant names from MS-RDPBCGR, fix broken field lengths, avoid magic values, add plenty of comments.


hean01-cendio:
Nice clean up!

PR #182, opened 2017-10-26
Local Mouse Cursor

joshudson:
Commit is from the issue85 branch; if you don't pull issue85 first it should cherry pick just fine


hean01-cendio:
Commit 87ccf3c cherry-picked into master.

PR #187, opened 2017-10-31
Added bootstrap to build instructions

hansfilipelo:
It is not mentioned within the README that one needs to run the script bootstrap / autoconf before running `./configure`. This adds one line to README and decreases level of entry for compiling the source.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ and 5 protocols.
 rdesktop uses a GNU-style build procedure.  Typically all that is necessary
 to install rdesktop is the following:
 
+	% ./bootstrap
 	% ./configure
 	% make
 	% make install


derfian:
Hey @hansfilipelo, thanks for the patch.

Technically speaking, we do already mention that you need to run `./bootstrap` if you're building from a snapshot, but looking at it again I can see that there is room for improvement.

I'd like to avoid telling people to run ./bootstrap when they don't need to. For instance, the README is also included in the source archives (as produced by `make dist`) together with a generated configure script. In this case, running bootstrap is not necessary.

Maybe we should just clarify the existing content under new headlines (I'm thinking "building from source archive" and "building from git") to make it more obvious which steps you need to take.

uglym8:
Or perhaps we could add a page to wiki dedicated to compiling rdesktop for various OSes/distros.

paulmenzel:
There is the following section. Isn’t that enough?

> ## Note for users building from source
>
> If you have retrieved a snapshot of the rdesktop source, you will first need to run ./bootstrap in order to generate the build infrastructure. This is not necessary for release versions of rdesktop.

uglym8:
@paulmenzel, yes,  I agree,  it seems a little bit excessive.
@hansfilipelo, @derfian, guys, do you have any objections against closing this issue?

derfian:
@uglym8 no objections from me

uglym8:
@hansfilipelo, thank you for your suggestion.
This time though we decided to move on without merging it.

Thanks again.

PR #193, opened 2017-11-09
Clean up undefined behavior in complicated and/or/shift expression

derfian:
Shifting negative values is technically undefined, and we can describe
what we want to do in ways that are easier to read.


derfian:
It appears that the new code isn't doing exactly what it used to do. I'll have to take another look at it.

hean01-cendio:
@derfian works like an charm ! ;)

derfian:
> It appears that the new code isn't doing exactly what it used to do. I'll have to take another look at it.

Scratch that. I blame @hean01-cendio for all the confusion. :D

PR #196, opened 2017-11-15
Remove device by its ID.

uglym8:
Using this function along with **rdpdr_send_client_device_list_announce()**
it's rather trivial to add support for dynamic devices in the future.

For example, you can detect smartcard reader insertion and pass it to
the TS (depending on your settings, of course). After removal you can
ask TS to remove it too.

Or you can implement dynamic disks support in such way that each new disk
is passed individually to the TS after insertion thus no need to supply
the disk path while starting rdesktop.

### DIFF
--- a/rdpdr.c
+++ b/rdpdr.c
@@ -276,6 +276,30 @@ announcedata_size()
 }
 
 static void
+rdpdr_send_client_device_list_remove(uint32 id)
+{
+	int i;
+	int size;
+	STREAM s;
+
+	/* MS-RDPEFS: 2.2.3.2 Client Drive Device List Remove (DR_DEVICELIST_REMOVE) */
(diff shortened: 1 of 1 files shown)


derfian:
So if I understand it right there's no code that actually uses this new `rdpdr_send_client_device_list_remove` function? I like the idea of having dynamic redirection of disks, smart card readers etc., but I'd like to avoid adding code that's not being used.

uglym8:
Yes.
You're absolutely right.
This function is not yet used anywhere.

I'm proposing to add it only because:
1) To let others know that such function exists as I had  hard time finding this PDU when I needed it.
2) To spark the interest in implementation of  the dynamic devices support:)

It's perfectly OK if you prefer not to merge it ATM.
It's enough for me to have this info around.

PR #200, opened 2017-11-23
Ignore object files.

alexandrevicenzi:
After building rdesktop on my computer git status showed a lot of ".o" files as untracked files.

Since ".o" files are not supposed to be added I think it's a good idea to hide them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ config.status
 configure
 rdesktop*.tar.gz
 aclocal.m4
+*.o


uglym8:
Never paid attention to this as I use  alias in .gitconfig:
````
ust = status -uno
````
but I guess we should definitely merge this.

LGTM.

@alexandrevicenzi thank you for your PR:)

hean01-cendio:
@marcosps :) nice find, I'm just like @uglym8 uses 'git status -uno' that hides untracked files.

PR #208, opened 2017-12-11
Update sv keymap to include neighbouring nordic characters

nospam3159:
As supported on virtually all other Swedish keyboard keymaps.


derfian:
The patch looks fine, and I was hoping to do a quick test and merge this, but I can't seem to write æøÆØ characters with this patch applied. No characters appear, although keyboard debug logging indicates that they're translated and sent to the server.

I'm confused. Maybe someone with more insight into the keymap/input side of rdesktop can help?

nospam3159:
Thanks for looking at and attempting to merge my patch!

For what it is worth, I have successfully been using this change since before I first posted it to rdesktop-devel in June. Yet only with rdesktop running on Linux connecting to VirtualBox hosts exposed through RDP. It is possible that I underestimate the complexity required to implement it properly for other environments.

Even though I lack the understanding of rdesktop internals, I could try to help debugging what happens in other setups. What kind of RDP server were you connecting to when it did not work @derfian ?

derfian:
> What kind of RDP server were you connecting to when it did not work @derfian ?

Windows 2016. I tried looking at the keyboard code in rdesktop, but it didn't alleviate my confusion.

nospam3159:
I've verified with a Windows server too now, and sure enough it does not work.

Did some more debugging and read the doc/keymapping.txt properly. Think I understand the issue and should have a new patch to post shortly.

nospam3159:
Having looked a few more minutes at the issue, I quite confident that the patch is correct as it is. It is however insufficient in delivering everything needed for Windows hosts. Applying the patch still improves functionality for us connecting from unix to unix.

Keystrokes are successfully sent if adding sequence-lines to the keymap file. For example adding "sequence oslash dead_diaeresis e" yields an ë for each press on ø, even on Windows. I can't figure out how to add a sequence that actually generates the desired characters though. Attempting combinations with dead_stroke, compose and Multi_key in them fails in different interesting ways. Maybe it is possible to fix, but maybe a fix requires patches on other levels.

I'm giving up at getting this to work with Windows.

nospam3159:
Branch rebased to current master. No other changes.

PR #211, opened 2017-12-13
Fix file sizes for redirected disks

uglym8:
Provide the full size for files during disk redirection.

We are not supplying full size info.
Check with file sizes bigger than 32 bits.

This affects not only display sizes in Explorer, but copy (and which is more severe appending data to the file too) operations are affected as well.

### DIFF
--- a/disk.c
+++ b/disk.c
@@ -5,6 +5,7 @@
    Copyright 2003-2011 Peter Astrand <astrand@cendio.se> for Cendio AB
    Copyright 2017 Henrik Andersson <hean01@cendio.se> for Cendio AB
    Copyright 2017 Karl Mikaelsson <derfian@cendio.se> for Cendio AB
+   Copyright 2017 Alexander Zakharov <uglym8@gmail.com>
 
    This program is free software: you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


uglym8:
You mean smth like  stream_write_64bit_length(STREAM out, off_t length)?

I guess it might come in handy at least for this particular case.
Wanna me to add it to this PR?
Btw, got a catchy name for this function?:)

hean01:
Nice, what I know is that `off_t` can be either 32 or 64 bits depending on compiler and specific flags. Adding and using a  `out_uint64_le()` stream macro with this PR would be nice, see stream.h for equivalent out_uint32_le() macro.

uglym8:
@hean01, @derfian: Done. Check, please, when you got a spare time.

hean01:
@uglym8 Looks nice, especially how clean the changes in disk.c got. My only concern is that you left out an implementation for big endian archs. For now, rdesktop tries to support big endian due to old SPARC and MIPS platform and I also know that ARM hardware platform can be configured to either little or big endian. With that I think we should continue support big endian archs, do you have any input on this topic ?

uglym8:
@hean01, yes, I should ask about BE support myself. 

IMHO, the only possible input here is to add the missing BE bits :)

CendioOssman:
AFAICT this is already endian agnostic?

uglym8:
Yes, we've already agreed upon this in the offsite conversation:)

I think the only part left is to remove compiler guard (L_ENDIAN) to make it available for both
endianess.

I'll fix it over the weekend.

uglym8:
Done.
Could not wait for weekend:)

PR #212, opened 2017-12-15
Rename functions that send Client Info PDU and handle Set Error Info PDU

derfian:
Being able to cross-reference our code with MS-RDPBCGR is a good thing. This PR renames the functions that handle the Client Info PDU and the Set Error Info PDU to fit with MS-RDPBCGR.

This PR also marks process_ts_set_error_info_pdu as static and removes it from proto.h. Nothing outside of rdp.c calls it.


hean01:
Looks good, what do you think about to set log level to warning upon an error ?

derfian:
I think "not an error" is also an error code.. and I'm not sure I want to emit a warning message based on that. I'll have to double-check though.

Issue #191, opened 2017-11-06
Enhancement: add a switch to NOT provide a password


posgnach:
Hello,

this is just a proposal. I have a shortcut on my desktop that runs rdesktop and connects to my laptop, which is running windows 10.
After commit 1aaafc80c037f57c3742ccb25b279467fcbf58b5 this shortcut broke, since rdesktop wants a password. I do not wish to write the password in the shortcut and I do not wish to run rdesktop from a terminal.
Can we opt out from providing a password? Windows 10 asks me to type it anyway.

uglym8:
Hello,

> After commit 1aaafc8 this shortcut broke, since rdesktop wants a password. I do not wish to write the password in the shortcut and I do not wish to run rdesktop from a terminal.

First of all, note that I don't advocate against your proposal in any way.
Having said this I can propose the following as a temporary workaround: you can provide the arbitrary password in your shortcut and later enter the right password when TS asks you to.

>Windows 10 asks me to type it anyway.

Why? 
I've just checked this and if the password is right you don't have to enter it again.

Sturmungandir:
> Why?
I've just checked this and if the password is right you don't have to enter it again.

I don't know. Could it be a feature of Windows 10 when it is part of a domain?
It will ask for a password again, through the usual Windows login screen.
It does not matter which password I provide to rdesktop.

uglym8:
> I don't know. Could it be a feature of Windows 10 when it is part of a domain?

Do you provide the domain (i.e. **-d**) info as well?

Sturmungandir:
Yes I do

uglym8:
Really strange.
I've got no W10 which is a part of a domain so unfortunately I'm unable to verify this:(

Sturmungandir:
Well, I don't really mind, since I am not allowed to write the true password anywhere, but I mind that rdesktop asks me for a password. As a workaround, I am supplying a fake password

hean01-cendio:
@Sturmungandir I'm not really in for adding a switch to disable the required password. Would your case be solved if you not forced to open a terminal ? Then you could provided a password from a pop-up UI prompt like gnome-ssh-askpass like this:

     /usr/libexec/openssh/gnome-ssh-askpass "RDP Remote password" | rdesktop -p - -u user -d DOMAIN server

Sturmungandir:
[...]

PR #216, opened 2018-01-02
Add option to omit password prompt

nospam3159:
Fix issue #191

Regression introduced by commit 1aaafc8.


trentasis:
Very interesting to recover thi feature that was available with previous version, in thinclient envjronment without this you can't use rdesktop

Thanks

gilbsfr24:
I agree, it is not necessary to add new parameters, it causes an upward compatibility break

trentasis:
Anybody can give a real example how to login to multiple windows 2012 without to enter two times the password?

Thanks

hean01-cendio:
Commit bc727e7 makes it possible to provide an empty password via -p argument that will set aside the password prompt and take the end user to remote login page.

PR #236, opened 2018-01-26
Refactoring of slow and fastpath handling and rdp_recv()

hean01-cendio:
This changes clarifies a chunk of code related to receiving
data and handle slow and fast path pdus.

### DIFF
--- a/rdp.c
+++ b/rdp.c
@@ -92,46 +92,76 @@ uint16 g_session_height;
 
 static void rdp_out_unistr(STREAM s, char *string, int len);
 
+/* reads a TS_SHARECONTROLHEADER from stream, returns True of there is
+   a PDU available otherwise False */
+static RD_BOOL
+rdp_ts_in_share_control_header(STREAM s, uint8 *type, uint16 *length)
+{
+	uint16 pdu_type;
+	uint16 pdu_source;
+
+	UNUSED(pdu_source);
+
+	in_uint16_le(s, *length);	/* totalLength */
+
+	/* If the totalLength field equals 0x8000, then the Share
+	   Control Header and any data that follows MAY be interpreted
+	   as a T.128 FlowPDU as described in [T128] section 8.5 (the
+	   ASN.1 structure definition is detailed in [T128] section
+	   9.1) and MUST be ignored.
+	*/
+	if (*length == 0x8000)
+	{
+		/* skip over this message in stream */
+		g_next_packet += 8;
+		return False;
+	}
+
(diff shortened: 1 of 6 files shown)


rdp.c
```diff
@@ -90,6 +90,37 @@ uint16 g_session_height;
 
 static void rdp_out_unistr(STREAM s, char *string, int len);
 
+/* reads a TS_SHARECONTROLHEADER from stream, returns True of there is
+   a PDU available otherwise False */
+static RD_BOOL
+rdp_ts_in_share_control_header(STREAM s, uint8 *type, uint16 *length)
+{
+	uint16 pdu_type;
+	uint16 pdu_source;
```
derfian: > rdp.c:99:9: warning: variable ‘pdu_source’ set but not used [-Wunused-but-set-variable]

I get compiler warnings about this variable. Do we need the information in `pdu_source` for anything? Should we just skip reading it?

rdp.c
```diff
@@ -98,38 +129,35 @@ rdp_recv(uint8 * type)
 	static STREAM rdp_s;
 	uint16 length, pdu_type;
```
derfian: > rdp.c:130:17: warning: unused variable ‘pdu_type’ [-Wunused-variable]

You removed all use of `pdu_type` from `rdp_recv`, so we can remove the local variable as well.

rdp.c
hean01-cendio: Ah ok, sometimes we have skipped those bytes, and sometimes we have tagged the variable like `UNUSED(pdu_source)`. I like the latter best due to clean readable code for packet parsing instead of `in_uint8s(s,2); ` which just skip over bytes

derfian: The updated patch doesn't look like it solved the "problem" - we're using `pdu_type` to set `*type` even though it's marked as unused. That doesn't make much sense to me.

We're also still writing a value to `pdu_source` but never using it for anything, even though it's marked as unused. The compiler is somehow more happy with this version, but I think it looks odd. I agree that just skipping the bytes is less than ideal.

hean01-cendio: >We're also still writing a value to pdu_source but never using it for anything, even though it's 
> marked as unused. The compiler is somehow more happy with this version, but I think it looks 
> odd. I agree that just skipping the bytes is less than ideal.
>

I there a third option for solution ? As I see it we have a choice between , read value to tagged unused variable or use skip byte function, what do you prefer, as mentioned earlier I prefer the method of tagging unused variables over skip x bytes for clarity.

Issue #233, opened 2018-01-12
SmartCard(warning): TS_SCardControl(), bogus smart card control code 0x00fd0028


trentasis:
Hi,

Detected with latest master branch an error when you try to insert some smartcard in the smartcard reader

SmartCard(warning): TS_SCardControl(), bogus smart card control code 0x00fd0028

trentasis:
error showed with keyboard cherryr SC K1000, and error appears when you insert some smartcard, not all, only smartcard that use safesign driver

SmartCard(warning): TS_SCardControl(), bogus smart card control code 0x00fd0028

uglym8:
Hi!
Sorry, but I doubt that this is the latest master as this warning was replaced in 87d8d12 (https://github.com/rdesktop/rdesktop/commit/87d8d123b81093ec8ce635e3c5009038e5a102f3#diff-92e289bf01abb35f501f99158d27604fL2143) (on Jan 26, 2017)
which was a year ago.

Does this error prevent you from using smartcard or just irritate you?

trentasis:
Hi,

error message is that appear in second message 
SmartCard(warning): TS_SCardControl(), bogus smart card control code 0x00fd0028

And appears when you insert smardcard and certificate is not showed, certificate not propagated and not usable 

What does it means this error? how to solve?

trentasis:
Sorry, what i atr and how can I obtain, smartcard is from safesign and error is reported only inserting smartcard  in the smartcard reader, no other action, If you givendetail what do you need I  can try to obtain

Thanks

trentasis:
Hi,

first thanks for your help.
1.- I don't know, pcsc_scan output:
./pcsc_scan
PC/SC device scanner
V 1.5.2 (c) 2001-2017, Ludovic Rousseau <ludovic.rousseau@free.fr>
Using reader plug'n play mechanism
Scanning present readers...
0: Cherry KC 1000 SC [KC 1000 SC] 00 00

Thu Jan 18 21:22:27 2018
 Reader 0: Cherry KC 1000 SC [KC 1000 SC] 00 00
  Card state: Card inserted,
  ATR: 3B BB 18 00 C0 10 31 FE 45 80 67 04 12 B0 03 03 00 00 81 05 3C

sh: ATR_analysis: not found
ATR_analysis '3B BB 18 00 C0 10 31 FE 45 80 67 04 12 B0 03 03 00 00 81 05 3C': Success

[...]

PR #238, opened 2018-01-27
Fix scard handling of PCSCv2 Part 10.

### LINKED ISSUE #75: Class 2/3 smart card reader features not working with pcsclite ccid driver
The changes in scard.c in commits 1ac2cd1, 91af7ba, b82d367 don't work with class 2 or class 3 smartcard readers that use the ccid driver from the pcsc-lite project. This driver uses control codes that overlap with the 49 << 16 used in Windows codes. In the output of GET_FEATURE_REQUEST these codes are translated to the Windows format. Then when using one of those codes, translation in the other direction is not done because the condition "(dwControlCode & 0xffff0000) == (49 << 16)" always results to false.

This might be fixed be adjusting the mask, but my opinion is that control codes returned by the reader driver should always be left alone, because only this way we can be sure that the driver will understand the codes sent back to it. Translating makes assumptions about the used bits that might not be valid.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/39356590-class-2-3-smart-card-reader-features-not-working-with-pcsclite-ccid-driver?utm_campaign=plugin&utm_content=tracker%2F19353203&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F19353203&utm_medium=issues&utm_source=github).
</bountysource-plugin>

uglym8:
At the moment after getting reply for **GET_FEATURE_REQUEST** we're modifying
codes for features presumably making them recognizable on Windows.
After we pass the modified values to the Windows, the Windows application uses these codes without
any modifications (as it should have as these codes are defined by driver itself and not by 
PC/SC).
After getting these codes (previously modified by us) we fail to recognize them as valid codes
and report them as bogus.
As a result no features from Part 10 of PC/SC v.2 will work correctly.

This PR should fix #75, and #233.

_**Example**_:
Windows application issues CM_IOCTL_GET_FEATURE_REQUEST and gets the following output
from the reader:
**06** 04 42 33 00 06 **07** 04 42 33 00 07 **0a** 04 42 33 00 0a **12** 04 42 33 00 12

This TLV means that the reader supports the following features with the corresponding control codes to
request them:

Code | Feature | Control Code
------------ | -------------|----------
**06**| FEATURE_VERIFY_PIN_DIRECT|0x42330006
**07**|FEATURE_MODIFY_PIN_DIRECT|0x42330007
**0a**|FEATURE_IFD_PIN_PROPERTIES|0x4233000a
**12**|FEATURE_GET_TLV_PROPERTIES|0x42330012

Windows application then issues **FEATURE_GET_TLV_PROPERTIES** using modified (**_by us_**) control code which instead of **0x42330012** becomes **0xfd0048**.
Having received this code we declare it bogus but send it anyway to the reader which obviously
does know nothing about such a code and reports back an error.

References:
1) http://pcscworkgroup.com/Download/Specifications/pcsc10_v2.02.01_sup.pdf
2) https://www.mail-archive.com/search?l=muscle@lists.musclecard.com&q=subject:%22%5C%5BMuscle%5C%5D+pcscd+support+for+Class+2+SPE+for+Cherry+SmartBoard+xx44%22&o=newest&f=1

### DIFF
--- a/scard.c
+++ b/scard.c
@@ -5,6 +5,7 @@
    Copyright 2010-2013 Pierre Ossman <ossman@cendio.se> for Cendio AB
    Copyright 2011-2017 Henrik Andersson <hean01@cendio.se> for Cendio AB
    Copyright 2017 Karl Mikaelsson <derfian@cendio.se> for Cendio AB
+   Copyright 2018 Alexander Zakharov <uglym8@gmail.com>
 
    This program is free software: you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
@@ -2185,6 +2186,7 @@ TS_SCardControl(STREAM in, STREAM out)
 	       (int) nOutBufferSize);
 
 	/* Is this a proper Windows smart card ioctl? */
+	/* FILE_DEVICE_SMARTCARD defined as 0x000031 in Winsmcrd.h  */
 	if ((dwControlCode & 0xffff0000) == (49 << 16))
 	{
 		/* Translate to local encoding */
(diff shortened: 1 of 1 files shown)


uglym8:
Verified SPE (Secure PIN entry) with ACOS smartcards and HP KUS1206 keyboard.

PR #241, opened 2018-01-29
Fix to compile with gcc 7.3.0 and possibly fix a security issue

kekePower:
Did a google search and found this.
https://stackoverflow.com/questions/4419293/warning-format-not-a-string-literal-and-no-format-arguments/4419319#4419319

With this little fix, rdesktop compiled again using gcc 7.3.0.

### DIFF
--- a/rdesktop.c
+++ b/rdesktop.c
@@ -517,7 +517,7 @@ read_password(char *password, int size)
 
 	if (tcgetattr(STDIN_FILENO, &tios) == 0)
 	{
-		fprintf(stderr, prompt);
+		fputs(prompt, stderr);
 		tios.c_lflag &= ~ECHO;
 		tcsetattr(STDIN_FILENO, TCSANOW, &tios);
 		istty = 1;


uglym8:
I don't think that it's necessary as **prompt** buffer is fixed (see lines 509-516).

If on the other hand this buffer (prompt) was not fixed (e.g. get data from the user) then your
changes will be not only desirable but actually mandatory.

kekePower:
@uglym8 
Do you have a better solution to how I can get rdesktop to compile using gcc 7.3.0?
Without this "fix", it errors out and stops and since I really don't know any C at all, I've no idea what this fix really does or if it introduces any regressions.

uglym8:
I guess you're right as some OSes may have  **-Werror=format-security** set
 in the default compiler options, so we can either use 

````C
fprintf(stderr, "%s", prompt); 
````
or the variant proposed by  you:

````C
fputs(prompt, stderr);
````
So, long story short: we should definitely merge this.
Thank you.

PR #242, opened 2018-01-29
Using gcc's fallthru to avoid compilation halting

kekePower:
Also did a google search on this and found out about FALLTHRU.
It may not be perfect, but at least I am able to compile rdesktop
on Mageia Cauldron with gcc 7.3.0.

Perhaps there are better ways to fix these issues, but I'm no coder.

### DIFF
--- a/bitmap.c
+++ b/bitmap.c
@@ -34,7 +34,7 @@
 #define CVAL2(p, v) { v = (*(p++)) << 8; v |= (*(p++)); }
 #endif /* L_ENDIAN */
 #else
-#define CVAL2(p, v) { v = (*((uint16*)p)); p += 2; }
+#define CVAL2(p, v) { v = (*((uint16*)p)); p += 2; } // FALLTHRU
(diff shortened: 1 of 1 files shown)


uglym8:
Thank you very much for reporting this.

I Saw the same error other day while compiling wireshark.
I'll wait until 7.3 arrives in my distro and then verify this PR.

hean01-cendio:
Using a comment with a special meaning to a specific compiler seems very fishy. I think we should refactor the switch cases to avoid fall throughs.

derfian:
I did some work on this a while back but I never got around to make a pull request.

https://github.com/derfian/rdesktop/commit/1cdf4e87cb6a1cd415b36c3531acc7b51a4069c8

I should probably just create a new PR for this - thoughts?

uglym8:
@derfian, yes, it'd be great.

uglym8:
@kekePower, thank you for your contribution.
This time we will stick to the refactoring (#245)

PR #249, opened 2018-03-13
Reconnect-related fixes

derfian:
This PR fixes two problems that were related to reconnecting.

* When connecting to a Windows 2008 Server, pressing Cancel or hitting Escape while entering credentials (on the server) would lead to a reconnect if there was a resize pending.
* When connecting to a RDS farm name (round robin DNS entry as per MS recommendations), reconnecting could connect you to a different server, leading to a login screen rather than your session.
* rdesktop never had any logging that told you which address you were connecting to.

### DIFF
--- a/tcp.c
+++ b/tcp.c
@@ -58,6 +58,13 @@
 #define STREAM_COUNT 1
 #endif
 
+#ifdef IPv6
+static struct addrinfo *g_server_address = NULL;
+#else
+struct sockaddr_in *g_server_address = NULL;
+#endif
+
+static char *g_last_server_name = NULL;
 static RD_BOOL g_ssl_initialized = False;
 static SSL *g_ssl = NULL;
 static SSL_CTX *g_ssl_ctx = NULL;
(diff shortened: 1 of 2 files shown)


tcp.c
```diff
@@ -425,38 +425,48 @@ tcp_connect(char *server)
...
-		return False;
+		if ((n = getaddrinfo(server, tcp_port_rdp_s, &hints, &res)))
+		{
+			logger(Core, Error, "tcp_connect(), getaddrinfo() failed: %s", gai_strerror(n));
+			return False;
+		}
 	}
 
-	ressave = res;
+	addr = res;
```
CendioOssman: If this is a round robin setup then `res` will contain all the addresses, so this change will not improve things. What we need to save is the specific `struct addrinfo` record that we managed to `connect()` to.

tcp.c
```diff
@@ -425,38 +425,48 @@ tcp_connect(char *server)
 	int i;
 
 #ifdef IPv6
-
 	int n;
-	struct addrinfo hints, *res, *ressave;
+	struct addrinfo hints, *addr;
+	static struct addrinfo *res = NULL;
```
CendioOssman: It's easy to miss static local variables, so perhaps it should be a static global?

tcp.c
```diff
@@ -423,6 +423,7 @@ tcp_connect(char *server)
 	socklen_t option_len;
 	uint32 option_value;
 	int i;
+	char buf[INET6_ADDRSTRLEN];
```
CendioOssman: Is this constant available on non-IPv6 systems? And do we care? :)

tcp.c
```diff
@@ -460,6 +461,22 @@ tcp_connect(char *server)
 		g_sock = socket(addr->ai_family, addr->ai_socktype, addr->ai_protocol);
 		if (!(g_sock < 0))
 		{
+			switch (addr->ai_family) {
+				case AF_INET:
```
CendioOssman: `getnameinfo()` allows you to get a string without having to care about the family.

tcp.c
```diff
@@ -460,6 +461,22 @@ tcp_connect(char *server)
...
+					logger(Core, Debug, "tcp_connect(), trying %s",
+					       inet_ntop(addr->ai_family,
+							 &((struct sockaddr_in6 *) addr->ai_addr)->sin6_addr,
+							 buf, sizeof(buf)));
+					break;
+				default:
+					return False;
+			}
 			if (connect(g_sock, addr->ai_addr, addr->ai_addrlen) == 0)
 				break;
```
CendioOssman: Perhaps explicitly log which succeeded? And why it fails for those that didn't?

tcp.c
derfian: I don't actually know. It's in POSIX, so I figured it was OK.

derfian:
Note to self: test the code before pushing it.

derfian:
> Note to self: test the code before pushing it.
Yeah, ignore the last commits. They're broken.

derfian:
I did some force-pushing to consolidate the original commits and fixes. 478a5cd ought to solve the server address problem.

derfian:
Hm.. not sure if it was my changes in 0d41492 or what, but reconnecting with CredSSP seems to disconnect me more often than not. Needs more investigation. Apart from that,  478a5cd seems solid.

derfian:
Aaaargh, spoke to soon. The IPv6 codepath errored out again.

derfian:
4e3cb44 fixes a couple of silly mistakes w.r.t. the IPv6 codepath - if g_server_address is used, don't memcpy it to itself, and don't free it either.

derfian:
0d41492 introduced a regression that breaks reconnect-based resize on Windows 2008 R2, leading to a disconnect instead of a reconnect.

> Protocol(debug): process_ts_set_error_info_pdu(), error info = 0
> [...]
> Window resize detected, reconnecting to new size 1060x802
> Disconnecting...

At this point, rdesktop will recognize that we got an error info and clear the resize flag. If we don't consider 0 a valid error value (the previous behavior), we will reconnect when the user presses escape at the server side login form.

CendioOssman:
I tested the last two commits and they seems to work well. I tested both the IPv6 and the legacy code paths. tcpdump shows just the one lookup from rdesktop (and an extra from libkrb during CredSSP handling).

I also noted this also had the benefit of reconnecting to the same server when we failed to negotiate CredSSP and had to fall back to plain TLS.

derfian:
I have fixed the remaining problems and cleaned up this PR into two distinct, complete, stand-alone commits.

> * When connecting to a Windows 2008 Server, pressing Cancel or hitting Escape while entering 
credentials (on the server) would lead to a reconnect if there was a resize pending.

8c9c52a

> * When connecting to a RDS farm name (round robin DNS entry as per MS recommendations), reconnecting could connect you to a different server, leading to a login screen rather than your session.
> * rdesktop never had any logging that told you which address you were connecting to.

d6c99bf

Issue #197, opened 2017-11-20
CredSSP not fully working


moobyfr:
Hi
My setup is:

- `2K12` server in a domain `SUB1.AD.DOMAIN.COM`
- linux used to auth users in `AD.DOMAIN.COM`
- the user is able to get a ticket with `kinit` 
When I make a RDP connexoin with rdesktop, I get a TGS for 
`TERMSRV/2K12.SUB1.AD.DOMAIN.COM@AD.DOMAIN.COM`
(`Connection established using CredSSP`)

on the server side, the auth seems to start, but it stops with "invalid password"
And the username prefilled is "AD.DOMAIN.COM\username"
this is strange, the normal login should be `AD\username` or `username@AD.DOMAIN.COM`

derfian:
It took me a little while to get there but I was able to reproduce this problem. The steps required to reproduce this scenario for me was to get a user ticket with kinit, run rdesktop but do *not* enter a password when prompted. Also, I wasn't able to reproduce if I got redirected to a session on another RDS server.

I seem to recall somewhere reading that CredSSP + Kerberos isn't used for the user authentication, just for authenticating the server, and that you still need a password (or smart card) to authenticate. I'd love to have someone else confirm that.

Which version of rdesktop are you using?

moobyfr:
I'm using 1.8.3 from debian repository stretch

uglym8:
I'm not an CredSSP expert but I think @derfian is talking about the following question by @hean01-cendio :https://social.msdn.microsoft.com/Forums/SqlServer/en-US/da074f0f-0887-4151-88ea-19a671ed91d9/is-it-possible-to-do-true-kerberos-sso-using-rdp-?forum=os_windowsprotocols

Also check #71, #72

hean01-cendio:
@uglym8 yes, that is true, CredSSP use of Kerberos is only for confidentiality and integrity to prevent MITM.

To simplify that, one should think 'is it safe to share my credentials with the remote end', kerberos ensures that.

hean01-cendio:
@moobyfr I agree, that it is strange the REALM is used as domain name prefix and not using a user principle name eg. user@REALM. I'll look though code to see if I can spot an error upon this topic.

What is the command line for your rdesktop connection, when this fails ?

moobyfr:
simply ``rdesktop tseserver`` while having a valid kerberos ticket

hean01-cendio:
[...]

PR #253, opened 2018-03-26
Net fixes

hean01-cendio:
Commit e94b384:

Change that DNS resolve only occurs once for server name provided by command line and for server name provided by server upon redirection. This fixes problem with reconnect triggered by dynamic session resize or network error, and DNS resolves different addresses due to round robin configuration.
This changes does also fix that regarding of what you enter on command line, CredSSP will use resolved FQDN for service principle name.

Commit 7d6466f:

Fixes the code that handles what should be done after rdp main loop exists. There are many variables that have impact on the decision. The code was refactored for clarity. This commit also adds a workaround for a difference between Windows 2008 Server and earlier and newer which fixes a problem with a reconnect triggered if one hit ESC on login screen.


cssp.c
```diff
@@ -115,18 +115,20 @@ cssp_gss_mech_available(gss_OID mech)
...
-
 	gss_OID type = (gss_OID) GSS_C_NT_HOSTBASED_SERVICE;
-	int size = (strlen(service_name) + 1 + strlen(server) + 1);
+
+	getnameinfo(ai->ai_addr, ai->ai_addrlen, fqdn, sizeof(fqdn), NULL, 0, 0);
+	size = (strlen(service_name) + 1 + strlen(fqdn) + 1);
 
 	output.value = malloc(size);
-	snprintf(output.value, size, "%s@%s", service_name, server);
+	snprintf(output.value, size, "%s@%s", service_name, fqdn);
```
CendioOssman: Won't this break systems where you don't have forward and reverse lookups in sync? `AI_CANONNAME` is probably a better idea, possibly with `getnameinfo()` as a fallback for ip addresses (if `AI_CANONNAME' doesn't do reverse lookup for you).

proto.h
```diff
@@ -19,6 +19,8 @@
 #ifndef RDESKTOP_PROTO_H
 #define RDESKTOP_PROTO_H
 
+#include <netdb.h>
```
CendioOssman: A small optimisation is to just declare `struct addrinfo;` here. You don't need the full header here.

tcp.c
```diff
@@ -418,91 +418,52 @@ tcp_tls_get_server_pubkey(STREAM s)
...
+	while (pai)
 	{
-		g_sock = socket(res->ai_family, res->ai_socktype, res->ai_protocol);
-		if (!(g_sock < 0))
+		g_sock = socket(pai->ai_family, pai->ai_socktype, pai->ai_protocol);
+		if (g_sock > 0)
 		{
-			if (connect(g_sock, res->ai_addr, res->ai_addrlen) == 0)
+			/* try connect to server */
+			res = connect(g_sock, pai->ai_addr, pai->ai_addrlen);
```
CendioOssman: The logging from the previous version was lost here. That would be nice to have when debugging.

tcp.c
```diff
@@ -418,91 +418,52 @@ tcp_tls_get_server_pubkey(STREAM s)
...
-		TCP_CLOSE(g_sock);
-		g_sock = -1;
+		logger(Core, Error, "tcp_connect(), unable to connect to %s",
+		       ai->ai_canonname);
 		return False;
 	}
 
-#endif /* IPv6 */
+	getnameinfo(pai->ai_addr,  pai->ai_addrlen, hostname, sizeof(hostname), NULL, 0, NI_NUMERICHOST);
+	logger(Core, Debug, "Connected to %s", hostname);
```
CendioOssman: I don't see it storing the result anywhere. It needs to remember which address it actually used or we'll get problems with reconnects after redirects.

tcp.c
```diff
@@ -561,6 +522,28 @@ tcp_get_address()
...
+	int res;
+	char service[16];
+	struct addrinfo hints={0};
+
+	snprintf(service, sizeof(service), "%d", port);
+
+	hints.ai_flags = AI_NUMERICSERV | AI_CANONNAME;
+	hints.ai_socktype = SOCK_STREAM;
+
+	res = getaddrinfo(address, service, &hints, result);
```
CendioOssman: Where do we free this? Or do we let it leak on each resolve?

cssp.c
hean01-cendio: As I understand and have verified, AI_CANONNAME would be the resolved name for the passed hostname to getaddrinfo(), this name can be a RDS farm name with round robin and this is not allowed to be used with a SPN which needs to be the actual hostname FQDN. So we can't use that.

proto.h
hean01-cendio: Will make it a forward declaration

cssp.c
hean01-cendio: Another problem exists, it resolves the first item in list and assume that this is the host connected to. Should be the entry int the list that a successful connection was made to. In most cases this will work, except when the first entry from a round robin is down and second one is used for connection.

tcp.c
hean01-cendio: I'll move the logging for a successful connection into the loop

tcp.c
hean01-cendio: I left it out because there will theoretically only be two allocations for a instance lifetime, one for commandline servername and one for redirection, both done in main()

cssp.c
CendioOssman: Shouldn't you add the farm SPN to each server in that case? I can find some people referring to using `setspn`, but this Microsoft blog seem to indicate that the connection broken can keep track of a common Kerberos identity:

https://cloudblogs.microsoft.com/enterprisemobility/2009/05/20/creating-kerberos-identity-for-rd-session-host-farms-part-i-using-the-remote-desktop-services-provider-for-windows-powershell/

tcp.c
CendioOssman: Only a problem if you want to run valgrind then. :)

cssp.c
hean01-cendio: That was a feature for only Windows 2008 that is deprecated. See first response to that blog post. When deploying a RDS session host a SPN is automatically created for the host which makes our approach to work out of the box against both 2008 and later.

tcp.c
hean01-cendio: I will fix that, this is also needed for CredSSP to resolve the correct name for the connection in case of other then first entry is used.

cssp.c
CendioOssman: Alright. And no issues with the security? Doing a reverse lookup for normal Kerberos makes it vulnerable to spoofing, so the same might be true here?

`AI_CANONNAME` seems to have issues as well, though: :/
http://k5wiki.kerberos.org/wiki/Projects/Trust_KDC-local_name_resolution
http://mailman.mit.edu/pipermail/kerberos/2011-July/017313.html

tcp.c
```diff
@@ -418,91 +419,71 @@ tcp_tls_get_server_pubkey(STREAM s)
...
+			}
 
-	struct hostent *nslookup;
-	struct sockaddr_in servaddr;
+			/* Failed to reconnect to previous connected host, lets fail */
+			if (connection_pai == pai)
+			{
+				pai = NULL;
+				break;
+			}
```
CendioOssman: Don't we want to keep this if we're trying to reconnect after a network error? We might get a few failed `connect()` before we get back in touch with the server.

hean01-cendio: `pai` is local variable and only used to instrument a connect failure further down, whats important here is the connection_pai which is the pointer to the connection to be made which only updates on successful connection.

moobyfr:
Fixes #197

derfian:
@trentasis I believe there were still somes issues with this PR, so #255 was created to deal with the immediate problems. I don't fully recall the differences any longer, maybe @hean01-cendio or @CendioOssman could clarify if there's anything left of value that wasn't addressed by #255.

We should probably rename or close this PR depending on what's left.

PR #258, opened 2018-04-03
Fix scard logon

### LINKED ISSUE #257: Opened transaction after smart card authentication in windows
This bug was mentioned in your mailing list at the <a href="https://sourceforge.net/p/rdesktop/mailman/message/34321666/">sourceforge</a>. More than that, issuer provided a patch which fix it. 
I applied this patch on the current version of master and it works fine. Can you merge it please?

uglym8:
I suggest to apply a patch from devel-list (sf) to fix #257

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -218,6 +218,10 @@ AS_IF([test "x$enable_smartcard" != "xno"], [
 	   exit 1
 	fi
 
+	AC_MSG_CHECKING([for PCSC-lite >= 1.6.0 (PnP/Notifications support)])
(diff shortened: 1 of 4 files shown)


uglym8:
I can't verify this patch atm, so I rely on the user's report (#257) of the successful testing.

uglym8:
Added 2 more commits:
1) Enable PnP notification in SCardGetStatusChange
2) Fix TS_SCardEstablishContext() implementation as the current approach causes Windows logon to fail detect smartcard reader insertion in a case when rdesktop was lauched without any attached readers and with enabled scard redirection (See #257)

uglym8:
The initial reporter has just verified that these patches do indeed fix his issue (#257)

Issue #263, opened 2018-05-01
How to set WM_CLASS on remoteapp?


relink2013:
When using rdesktop to connect to remote apps how can I set the WM_CLASS of the window? This can be done in freerdp, and its nice because it allows you to easily make desktop launchers for each remoteapp.

uglym8:
Hi,
WM_CLASS and WM_NAME are set to "rdesktop" by default.
ATM, there's no option to set WM_CLASS.
You can check my temp branch (https://github.com/uglym8/rdesktop/tree/add_wm_class) to verify that newly added functionality (use **-w** option to set a window class) is what you really need.

uglym8:
Added PR #265.
If there will be no objections from other team members it'll be soon merged into master.

uglym8:
@hean01-cendio, I had a feeling that this is not that easy:)

Never tried to use Seamless RDP myself.
I'll try to dive deeper using your hints about how connection sharing works.

PR #265, opened 2018-05-10
Add option to set WM_CLASS

uglym8:
Done per user request in #263


ThomasAdam:
Hi @uglym8,

It's very unusual for the end-user to set/override the WM_CLASS property.  This is because of how further properties are used by window managers, where there's assumptions about this string (such as the class name starting with a capital letter).

What are you using this proposed change for?

Incidentally, if you could, you should double-check that the default WM_CLASS property is set to `Rdesktop` and not `rdesktop` as this is in violation of the ICCCM2/EWMH, although I appreciate not part of your original PR.

PR #271, opened 2018-06-18
(README): add instructions to install autoconf

goldylucks:
I got an error `./bootstrap: 3: ./bootstrap: autoreconf: not found` when trying to install.

If you prefer not to add this line to the "default" installing steps, maybe add it below?

something like "if you run into `autoreconf: not found` error, follow this step", or something similar


hean01-cendio:
I have created an issue #272 for adding information about required and optional tools / libraries for building rdesktop.

I will close this PR with the reason that the changes is incomplete with issue #272 and that it is Ubuntu specific.

Issue #284, opened 2018-10-18
Add TLS 1.1 and 1.2 support


gpatel-fr:
Hello

trying to connect with current rdesktop to a Windows server where Tls 1.0 has been disabled leads to

Failed to connect, CredSSP required by server.

Since adding TLS 1.1 and 1.2 support to rdesktop is not exactly rocket science I could try to send a PR if there is interest.

uglym8:
Hi!

Thanks for the information.

Yes, please, send a PR.
I've glanced through your patch and it seems OK to me.

I'd like to clarify one question though:
 is there a way for server to report what TLS version it supports?

It'll eliminate the need for the user to supply the CLI option indicating the TLS version to use.

TIA.
Best regards.

gpatel-fr:
typically it works like this
Client says its highest protocol version
Server answers with its own highest version
Client either picks server version or deny access

The deny access is not a theoretical issue, it may be mandatory that a client refuses to downgrade the conection for security reason. So an ideal design could be IMO
-no tls command line option: start with TLS 1.2 and downgrade until a connection is established
- a tls command line is specified: try to connect with this tls version and if it fails deny the connection

with this patch neither of this is done unfortunately. if 1.2 is specified and the server allows 1.0 and not secured connection rdesktop will downgrade to unsecured, not to tls 1.0. The actual downgrade to insecured has nothing to do with my patch it's probably an inheritance of pas times where Windows servers were connected to in clear, before server 2003 I guess

uglym8:
ok, thanks for the update.
Really appreciated that.

PR #285, opened 2018-10-20
Add TLS 1.1 and 1.2 support

gpatel-fr:
see issue #284

### DIFF
--- a/doc/rdesktop.1
+++ b/doc/rdesktop.1
@@ -116,6 +116,10 @@ connection sharing feature which spawns another notepad in the current connectio
 specified server and then exit.
 
 .TP
+.BR "-V <tls version>"
+Set the Transport Level Security (also known as SSL) Version used.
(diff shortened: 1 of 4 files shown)


uglym8:
First of all, I'd like to thank you for your time and good job on this PR.
It looks perfect.

I think I have to request a minor change from you in your PR though: 
I guess it'd be really convenient to inform the user about "-V" option in case of a failure.

Something like this:
If rdesktop fails to connect to the server and we believe the culprit is the TLS version we're using to connect to, then, we should output the following message (choose any wording you like as it's just an example):
Failed to connect to the server. Please use -V option to set TLS version.

What do you think?

uglym8:
Great:)
Put this change into this PR and I'll merge it.
Thanks.

gpatel-fr:
Done, added a few comment fixes I forgot in the first pass.
Thanks for taking the time to review it.

PR #289, opened 2018-11-14
Allow user to ungrab the keyboard by pressing certain key combination

### LINKED ISSUE #22: Prevent grabbing of certain key combinations
I'd like to prevent certain key combinations to be sent to the server. Problem is that rdesktop grabs ctrl-alt-arrow keys so I'm unable to navigate workspaces on my linux machine - as soon as I switch to a workspace with rdesktop I get stuck and have to use the mouse.

I've searched the net for an answer but can't find any. The best I found is a patch http://www.artificialworlds.net/blog/2011/07/ - but this requires ctrl-alt-arrow to be pressed twice to work. 

Would it be possible (or is it already possible? how?) to extend the keymap file to indicate that certain key cominations should not be passed to the server? Using -K is not an option since then you loose alt-tab, alt-F4...

I've also tried xfreerdp and remmina and both lack this functionality - or it works poorly.

uglym8:
This should fix #22.

You could use new option "**-J**' to set the key combination to initiate a keyboard ungrab.
Default is "**ctrl_alt**".

Available mods:
 - **supers** - both Super key pressed and released
- same for  **alts**, **ctrls** and **shifts**

- **super** - any of Super keys (either left or right)
-  same for **alt**, **ctrl** and **shift**

- **lctrl** or **rctrl** - left or right Control

For example:
**alts** - Pressing and releasing both Alts
**alt_rctrl** - Pressing and releasing any Alt and right Control
**lsuper_ralt** - Pressing and releasing left Super and right Alt

Tested on:
i3wm
Xfce
KDE
Gnome
MATE

Sure, could be some rough edges, but basic functionality should just work.

@hean01-cendio, if you get a chance please check whether this is really what you talked about in https://github.com/rdesktop/rdesktop/issues/22#issuecomment-255704219

### DIFF
--- a/xwin.c
+++ b/xwin.c
@@ -82,6 +82,8 @@ static int g_x_socket;
 static Screen *g_screen;
 Window g_wnd;
 
+static RD_BOOL g_has_wm = False;
+
 RD_BOOL g_dynamic_session_resize = True;
 
 /* These are the last known window sizes. They are updated whenever the window size is changed. */
@@ -1893,6 +1895,67 @@ set_wm_client_machine(Display * dpy, Window win)
 	XSetWMClientMachine(dpy, win, &tp);
 }
 
+RD_BOOL is_wm_active(void)
+{
+	Atom prop, actual_type;
+	int actual_fmt;
+	unsigned long nitems, bytes_left;
+	unsigned char *data;
+	Window wid;
+
+	prop = XInternAtom(g_display, "_NET_SUPPORTING_WM_CHECK", True);
+
+	if (prop == None) return False;
+
+	if (XGetWindowProperty(g_display, DefaultRootWindow(g_display), prop, 0, 1, False,
+				XA_WINDOW, &actual_type, &actual_fmt, &nitems, &bytes_left, &data) != Success) {
+		return False;
+	}
+
+	if (!nitems) {
+		XFree(data);
+		return False;
+	}
+
+	wid = ((Window *)data)[0];
+	XFree(data);
+
+	if (XGetWindowProperty(g_display, wid, prop, 0, 1, False,
+				XA_WINDOW, &actual_type, &actual_fmt, &nitems, &bytes_left, &data) != Success) {
+		return False;
+	}
+
+	if (!nitems) {
+		XFree(data);
+		return False;
+	}
+
+	if (wid != ((Window *)data)[0]) {
+		XFree(data);
+		return False;
+	}
+
+	XFree(data);
+
+	/* Just for the curious minds */
+	prop = XInternAtom(g_display, "_NET_WM_NAME", True);
+
+	if (prop == None) return False;
+
+
+	if (XGetWindowProperty(g_display, wid, prop, 0, 1, False,
+				AnyPropertyType, &actual_type, &actual_fmt, &nitems, &bytes_left, &data) == Success) {
+		if (nitems) {
+			logger(GUI, Verbose, "%s(): WM name: %s", __func__, data);
+		}
+
+		XFree(data);
+	}
+
+	return True;
+}
+
+
 /* Initialize the UI. This is done once per process. */
 RD_BOOL
 ui_init(void)
@@ -1970,6 +2033,8 @@ ui_init(void)
 		seamless_init();
 	}
 
+	g_has_wm = is_wm_active();
+
 	return True;
 }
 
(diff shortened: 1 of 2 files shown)


derfian:
@uglym8 if you don't mind, I'll split the EWMH fullscreen fixes (nice!) to a separate PR, there shouldn't be any opposition to merging these right away. I'll take a look at grab/ungrab hotkeys after that.

uglym8:
> @uglym8 if you don't mind, I'll split the EWMH fullscreen fixes (nice!) to a separate PR, there shouldn't be any opposition to merging these right away. I'll take a look at grab/ungrab hotkeys after that.

Sure.
I guess it's the correct approach as we might encounter some issues with grabbing/ungrabbing hotkeys (though hope that we won't:))

xwin.c
```diff
@@ -1895,6 +1959,146 @@ set_wm_client_machine(Display * dpy, Window win)
...
+	long bv;
+
+	bv = parse_mod_request(str);
+
+	if (bv != -1) {
+		g_grab_ctx.mod_supplied = bv;
+		g_grab_ctx.mod_allowed = g_grab_ctx.mod_supplied;
+		return 0;
+	} else {
+		logger(Keyboard, Warning, "%s(): Failed to parse supplied ungrab key combo (%s). Setting to default (any_ctrl + any_alt)", __func__, str);
```
derfian: Would it be possible to have this say `ctrl_alt` instead of `any_ctrl + any_alt` to prevent confusion about the key combo format?

xwin.c
```diff
@@ -2596,6 +2800,123 @@ handle_button_event(XEvent xevent, RD_BOOL down)
...
+
+			logger(Keyboard, Debug, "KeyRelease after: mod_allowed = 0x%lx mod_supplied = 0x%lx mod_in_progress = 0x%lx hit_ungrab_combo = %ld\n", ctx->mod_allowed, ctx->mod_supplied, ctx->mod_in_progress, ctx->hit_ungrab_combo);
+			break;
+		default:
+			break;
+	}
+
+	return 0;
+}
+
```
derfian: Ideally, I would have preferred to see this split into two more specific functions that would more clearly communicate their purpose (bikeshedding: `ungrab_save_keypress`/`ungrab_modifiers_released`), but I'm not going to be picky about it.

xwin.c
```diff
@@ -2596,6 +2800,123 @@ handle_button_event(XEvent xevent, RD_BOOL down)
...
+int is_ungrab_required(struct grab_combo_ctx *ctx, XEvent *ev)
+{
+	int i;
+	KeySym ks;
+
+	switch (ev->type) {
+		case KeyPress:
+			ks = XLookupKeysym((XKeyEvent *)ev, 0);
+
+			logger(Keyboard, Debug, "KeyPress before: mod_allowed = 0x%lx mod_supplied = 0x%lx mod_in_progress = 0x%lx hit_ungrab_combo = %ld\n", ctx->mod_allowed, ctx->mod_supplied, ctx->mod_in_progress, ctx->hit_ungrab_combo);
```
derfian: Please remove the newlines in the logger messages.

derfian: There's many instances of this so feel free to do it in all of the new code.

uglym8:
@derfian, first of all, thank you for the review.

I agree with all  your remarks.
I'll fix'em a little bit later.

> This doesn't seem to work on Gnome on Fedora 29 (haven't tried other platforms yet). The rdesktop window immediately gets a FocusIn event after ungrabbing the keyboard, leading to a new XGrabKeyboard.

Just verified myself with **Fedora 27** and **Gnome Classic** (after fixing the fullscreen handling,  with modern Gnome I don't have to ungrab keys to switch workspaces) using **ugrab_kbd** branch from my repository:

1. run rdesktop in fullscreen (-f)  and try to use Workspace Switching key combo (Super + PgUp/PgDown).
I don't use Ctrl+Alt+Up/Down as I use QEMU for running Fedora VM and this combination (Ctrl + Alt) is reserver to grab host pointer.

Result: you can't switch workspaces as rdesktop stil holds the grab

Note: to verify that workspaces switching is really working, you can do the following (I'll illustrate it with active rdeskop to make the picture more tense):

1. Press Ctrl+Alt+Entrer to leave fullscreen mode in rdesktop.
2. Click on the top system panel
3. Use Super + PgUp/PgDown to switch workspace

Result: Yes, you can switch workspaces now

2. run rdesktop in fullscreen (-f) and provide  **-J alt_shift** and try to use Workspace Switching key combo (Super + PgUp/PgDown).
Result: you still can't to switch workspaces as the rdesktop holds the grab.

Now, press Alt+Shift, and then try to use Super+PgUp/PgDown to switch the workspaces.

Result: Now you can do it.

You may even not supply the -J option and rely on the default: ctrl+alt behaviour to achieve the same goal.

As a side note, I've got several people who are using  this feature on the daily basis, so I consider this to be at least "working on some distros + DEs combinations":)

PR #298, opened 2019-01-17
Replace OpenSSL with GnuTLS Nettle, libtasn1

hean01-cendio:
This is a work branch of GnuTLS provided  by Alexander's PR #137 

- Added our own tdb cache for storing pubkeys, this solves the issue
  were several keys for same host was saved upon update of key.

- Added routine for handling a dialog with a choice input from user which
  easily could be converted to fl_choice() if wanted..


hean01-cendio:
gnutls_base64_decode2 / gnutls_base64_encode2 requires gnutls 3.60
gnutls_store_pubkey and alike requires gnutls v3.0

I think we could be ok with setting a requirement of gnutls 3.0 that was released 2011 in comparation with 3.60 that is released 2017-08-21.

that means, ditch using base64 encode/decode from gnutls api and use nettlw library instead

hean01-cendio:
@uglym8, not yet, i need to recap on a few things, i will summarize theme here for further discussion

hean01-cendio:
@uglym8 

To summarize the recent changes:

The peer certificate is first verified against system CA trust store. This way a system administrator can deploy a trusted CA certificate to make the certificate verification succeed.  If the certificate is invalid in some way; expired, revoked etc. a fallback is made to the exception mechanism. The exception mechanism provides a way for the user to add an exception for the certificate pinned to the host this is done per user just like when adding an exception when connecting to a web server with and invalid certificate using firefox. This exception mechanis is very needed due to standard installed RDS services is using a self-signed certificates.

There are still  issues to resolve with the current implementation, A RDS broker sends a redirect using an IP address, this means that even if a certificate issuer is trusted by the system, the verification will fail due to hostname / DN mismatch and user needs to add an exception.

I need to investigate how Microsoft client works in this case, does it suffer from same problem, or are there other requirements for example; the hostname entered by user is used in both certificate verification (first connection and redirected connection) and certificate requires to have subject alternate names specifying a virtual name for the farm (the one the user entered) and the host.

hean01-cendio:
> There are still issues to resolve with the current implementation, A RDS broker sends a redirect using > an IP address, this means that even if a certificate issuer is trusted by the system, the verification will > fail due to hostname / DN mismatch and user needs to add an exception.

This issue is solved with commit d12204b424c82253d9da718fddb1b1a1d945d3a8

hean01-cendio:
@uglym8 I need some comments / reasoning from you about the code blocks that are guarded by checking version of gnutls. I want to decide on required gnutls version and if we really need those code blocks we should bump required version and remove the guard, or just drop them.

 - [ ] Test against RDS cluster without trusted CA certificate
 - [ ] Test against RDS cluster with trusted CA certificate
 - [ ] Describe the reason for invalidation of certificate in the exception dialog
 - [ ] Go through the text shown to the end-user and make them very clear for
       easy understanding
 - [ ] Go through logging messages and their levels

uglym8:
> @uglym8 I need some comments / reasoning from you about the code blocks that are guarded by checking version of gnutls. I want to decide on required gnutls version and if we really need those code blocks we should bump required version and remove the guard, or just drop them.

While deciding whether we should get rid of the guards and simply bump the required gnutls version we should consider the following factors as well:

according to https://repology.org/metapackage/gnutls/versions there is at least one major distribution ( CentOS 6) which doesn't have a decent library version.
CentOS 6 will be EOL'ed on November, 30, 2020 (https://wiki.centos.org/About/Product).
So bumping gnutls version will effectively make rdesktop unusable on this distro.

I'd prefer to leave GNUTLS guards where they are now as the codeblocks which they are guarding are rather small and trivial.

Nevertheless we could get rid of gnutls guars in ssl.c  and leave comments stating smth like "use gnutls_x509_crt_get_pk_oid() for getting public key algorithm after bumping the gnutls requirement to > 3.5.0." or leave no comments at all assuming that this PR's very comment will effectively serve as its placeholder.

hean01-cendio:
> according to https://repology.org/metapackage/gnutls/versions there is at least one major distribution ( CentOS 6) which doesn't have a decent library version.
CentOS 6 will be EOL'ed on November, 30, 2020 (https://wiki.centos.org/About/Product).
So bumping gnutls version will effectively make rdesktop unusable on this distro.

CentOS 6 / RHEL6 (6.10 final) is shipped with gnutls 2.12.23 which means that we can not build current GnuTLS branch on that platform due to we use API >= 3.0.0.

However one can bring GnuTLS 3.5.19 from EPEL to these platforms.

uglym8:
>CentOS 6 / RHEL6 (6.10 final) is shipped with gnutls 2.12.23 which means that we can not build current GnuTLS branch on that platform due to we use API >= 3.0.0.

Ah, completely forgot about this.
My mistake. 
Sorry.

So we can ditch the support for GnuTLS < 3.0.0 or alternatively we could write our own 
gnutls_store_pubkey/gnu_verify_stored_pubkey() etc

But having 
> However one can bring GnuTLS 3.5.19 from EPEL to these platforms.

I guess  you could disregard my comment related to CentOS 6 support.

hean01-cendio:
> Nevertheless we could get rid of gnutls guars in ssl.c and leave comments stating smth like "use gnutls_x509_crt_get_pk_oid() for getting public key algorithm after bumping the gnutls requirement to > 3.5.0." or leave no comments at all assuming that this PR's very comment will effectively serve as its placeholder.

It seems like the algorithm check in the guard in ssl.c is redundant with the algorithm check in libtasn_read_cert_pk_parameters which is used if check_pk_algo arugment is 1.

Why can't we always rely on check_pk_algo = 1 and drop that redundant guarded block?

uglym8:
>> Nevertheless we could get rid of gnutls guards in ssl.c and leave comments stating smth like "use gnutls_x509_crt_get_pk_oid() for getting public key algorithm after bumping the gnutls requirement to > 3.5.0." or leave no comments at all assuming that this PR's very comment will effectively serve as its placeholder.

>It seems like the algorithm check in the guard in ssl.c is redundant with the algorithm check in libtasn_read_cert_pk_parameters which is used if check_pk_algo arugment is 1.

> Why can't we always rely on check_pk_algo = 1 and drop that redundant guarded block?

Sure, we can.

I just wanted to employ _**gnutls_x509_crt_get_pk_oid()**_ to get public key algorithm and later (during the first Travis build actually) discovered that it was introduced only in **3.5.0**.

This being said we can either remove this guard it completely and employ libtasn1 for this (as it is done in check_pk_algo=1 clause) or bump GnuTLS to 3.5.0 and get rid of this (check_pk_algo) clause in _**libtasn_read_cert_pk_parameters()**_ .

But  I reckon we could simply get rid of the guard and remove 'check_pk_algo' clause in **_libtasn_read_cert_pk_parameters()_** simultaneously.

This way we will always use _**libtasn_read_cert_pk_parameters()**_ to get  public key algorithm.

hean01-cendio:
About base requirment, I'm very comfortable about raising requirments to >=3.0.0 there is 3 guards in code that could be remove if we raised up to >=3.2.0. I'm pretty comfort about raising up to 3.2.0, looking https://repology.org/metapackage/gnutls/versions, shows that there are only very few distributions (counted 8 of 254) that matches >=3.0.0 && < 3.2.0.

uglym8:
>About base requirment, I'm very comfortable about raising  requirments to >=3.0.0 there is 3 guards in code that could be remove  if we raised up to >=3.2.0. I'm pretty comfort about raising up to  3.2.0, looking https://repology.org/metapackage/gnutls/versions, shows that there are only very few distributions (counted 8 of 254) that matches >=3.0.0 && < 3.2.0.

I'm also positive that we could safely get rid of these guards then.

Rakasa:
What version of libtasn1 is needed?

running ./configure, I get:
`rdesktop requires libtasn1. Please install the dependency`

However, trying to install the library, I get
`$ sudo dnf install libtasn1
Last metadata expiration check: 0:00:31 ago on Tue 12 Feb 2019 09:29:18 AM EST.
Package libtasn1-4.13-5.fc29.x86_64 is already installed.
Package libtasn1-4.13-5.fc29.i686 is already installed.
Dependencies resolved.
Nothing to do.
Complete!
`

~ Rakasa

ps: my apologies if I posted this in the wrong place...

uglym8:
Hi @Rakasa,
Create new issue, please.

PR #305, opened 2019-01-28
Fix checking the length of remaining data in stream for very compact orders.

XiaopengZHOU:


derfian:
Thanks, we're looking into it.

derfian:
I've added a refactoring of the length handling to make it a little more understandable.
Thanks for your contribution!

PR #306, opened 2019-01-28
Change the way we do fullscreen

derfian:
Based on work from @uglym8 on #289.

This implements support for fullscreen through Extended Window Manager Hints (EWMH).

### DIFF
--- a/xwin.c
+++ b/xwin.c
@@ -86,6 +86,8 @@ static int g_x_socket;
 static Screen *g_screen;
 Window g_wnd;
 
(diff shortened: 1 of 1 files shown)


derfian:
Looks great!

I was curious if we still need to recreate the X11 window on toggling fullscreen if we use this `_NET_WM_STATE_FULLSCREEN` property, but after glancing at the code I was quickly reminded of how much of a mess xwin.c really is, so I'm happy with this.

Issue #295, opened 2018-12-25
redirection local disk can not generate single file size > 4.3G


andysendo:
debian9 stable, rdesktop 1.8.3, connet to win7 server(local host kvm)

redirection local disk for file sharing can not create single file > 4.3G.

not sure it's rdesktop bug or hard code limit? or RDP protocol limit?

gpatel-fr:
rdesktop limitation I'm afraid. When it was first written I'm sure that Windows server were limited (you can find posts all around the net about limitations at 2 Gb then 4 Gb), so there is somewhere in the code a 32 bits counter and when it goes back to 0 the file beginning get overwritten.

Windows server are fully capable to go beyond 4 Gb  now, and rdesktop could be modified to go there too - I did a test in my own forked repo and it seems to work all right. 

I'm not sure if there is still a contributor to examine and merge a PR, let alone do a new release. Well, let's wait a bit to see if there is anyone still alive here.

andysendo:
thanks for reply,  hope new release will fix it.

uglym8:
Hi,
sorry for the late reply.

It has already been fixed in #211.
At least I hope it has been:)

Please try to build master and check whether this error still persists.

Thank you.

gpatel-fr:
Alexander, I did not even notice your PR :-) So I assumed it did not work, tried master and indeed it did not work for me. You can check out the big file branch in my fork, it worked for me (transfer of a 5 Gb file in both senses, result checked with md5sum)

uglym8:
OK, I'll definitely check it.
Maybe I missed something as it obviously the case:)

Just to clarify the symptoms:
You can't copy file with size exceeding 4.3 Gb either from Windows or to Windows.
Right?

gpatel-fr:
Alexander, the file is copied but truncated at 4 Gb. The part over 4 Gb is written over the beginning.
With master i tested only in the Windows -> Linux direction, don't think I tested in the other way.
I tested my branch in both directions.

uglym8:
@gpatel-fr, I've checked your modifications.

Yes, you're right, my changes from #211 don't fix the problem (better say they fix different problem).
According to MS-RDPEFS doc (2.2.1.4.3(4)) offsets really are 8 bytes long.

Please, make a PR and I'll be happy to merge it.

Thank you.

uglym8:
Closing this as resolved by #307.
Thanks @gpatel-fr , @derfian

PR #307, opened 2019-01-29
Fix device redirection offsets

derfian:
These changes makes sure we read all 64 bits of offset from device redirection read and writes, avoiding problems with files larger than 4 GB. A very brief test run shows that it can transfer a 5 GB file to and from a Windows server without messing up a the file contents past 4GB.

The big endian codepaths for reading/writing 64-bit values to streams are untested.

This should solve issue #295.

### DIFF
--- a/disk.c
+++ b/disk.c
@@ -582,7 +582,7 @@ disk_close(RD_NTHANDLE handle)
 }
 
 static RD_NTSTATUS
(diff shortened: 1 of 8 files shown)


derfian:
Force-pushed an updated commit with fixed logging and co-author credits to @gpatel-fr.

PR #309, opened 2019-02-01
fix heap overflow in libtasn_read_cert_pk_parameters

markusbeth:
use correct size to memcpy() the exponent data in
libtasn_read_cert_pk_parameters


uglym8:
@markusbeth, thanks for catching this copy-paste error.
Yes, it will lead to overflowing the buffer allocated for public exponent.

PR #311, opened 2019-02-05
Fix memory leaks

markusbeth:
Variables of type mpz_t must be cleared with mpz_clear().
Certificates initialized with gnutls_x509_crt_init() must be freed with gnutls_x509_crt_deinit().
The return values of XGetAtomName() have to be freed with XFree().


CendioOssman:
Looks good. Thanks.

PR #318, opened 2019-02-26
Rdp2tcp

Sec42:
I updated the patch from http://rdp2tcp.sourceforge.net/ to the current git version.

Would be nice to have support for this in the main rdesktop client.


CendioOssman:
I'm afraid this is not something that is part of our goal at this time. We prefer extensions to be in the main code, and we'd like to extensions that are commonly available on servers.

Issue #333, opened 2019-06-10
v1.8.6 and master getting overrun assert failure connecting to xrdp


amworsley:
Trying to connect to the Debian server running xrdp 0.9.9-1 (Buster) with versions newer than 1.8.4 get an unexpected stream overrun failure.

My debugging shows the older 1.8.4 code had many 00 padded fields and the newer one shows the packet has no-zero's and the length is way too big. 

See below for details

Compiling the 1.8.6 package with Debian and running with debug I get:
<pre>
RDESKTOP_DEBUG=All ./rdesktop    azza
Connection established using plain RDP.
ERROR: rdp.c:128: rdp_recv(), unexpected stream overrun0000 03 00 01 b5 02 f0 80 68 00 01 03 eb 70 81 a6 08 .......h....p...
0010 00 00 00 16 cf 1f 5e f8 58 db c7 8c cb 40 34 5b ......^.X....@4[
0020 08 b2 3b a8 79 da dd b3 1c 50 d5 aa b0 3e a4 01 ..;.y....P...>..
0030 fd 2e d8 e5 06 05 63 92 ca fe 87 db 41 09 7b be ......c.....A.{.
0040 3c f5 4f d8 d6 de bb b1 81 fe c0 2a d3 e5 53 08 <.O........*..S.
0050 16 ee f0 74 b1 01 a8 be 6f 9d 8d b2 96 73 7f b2 ...t....o....s..
0060 32 f5 7b 0d 78 30 f3 24 49 91 52 79 30 e2 a3 06 2.{.x0.$I.Ry0...
0070 ef a8 96 46 ae 54 dc ae 21 ce 7b a0 32 9f 2a 29 ...F.T..!.{.2.*)
0080 d9 4c fc b1 da 37 19 46 b8 3b c1 c7 5a 2b f3 ba .L...7.F.;..Z+..
0090 96 fb d6 2e 44 62 73 b1 2e d4 7e a1 77 ac 91 2e ....Dbs...~.w...
00a0 cd 22 a0 33 fb 8e 26 d1 0e b6 d7 2c 40 7f e0 d2 .".3..&....,@...
00b0 ba 10 ce d1 a0 cf 9a c4 7f 09 95 28 6d 73 a3 fd ...........(ms..
00c0 00 28 e8 9e bc cb a1 d2 5b 98 2a bd d5 b0 78 00 .(......[.*...x.
[...]

PR #334, opened 2019-06-11
sec_decrypt() the correct amount of data

markusbeth:
Save the correct amount of data to sec_decrypt() because after
inout_uint8p() the macro s_remaining(s) will find nothing left and
therefore calls sec_decrypt() with a length of 0.

This fixes connection problems with VirtualBox and possibly issue #333.


CendioOssman:
Good catch. Thanks.

Issue #342, opened 2019-08-04
Can create session, but can't get back into it.


Billybangleballs:
I upgraded to **rdesktop:armhf 1.8.6-2\~deb9u1** from **rdesktop:armhf 1.8.4-1\~deb9u1** (which worked fine btw), and now, although I can start a session on the remote computer, if I disconnect for any reason, I can't reconnect. I get an error msg :-

```
ERROR: rdp.c:140: rdp_recv(), unexpected stream overrun0000 03 00 00 1e 02 f0 80 68 00 01 03 eb 70 10 08 00 .......h....p...
0010 02 03 c7 3e 63 ee 58 8c 45 d9 04 00 16 00       ...>c.X.E.....
```

I also can't connect to existing sessions started by other clients, I get the login screen, but it bombs out as soon as I put in the password.

If I could unapt-get upgrade, I would, but I'm not that technical.
Is this something that might be fixable, or have I broken it completely?

Billybangleballs:
So after much googling, I figured out how to downgrade back to where it worked, but obviously this hasn't fixed the problem with rdesktop:armhf 1.8.6-2\~deb9u1 but it has fixed my immediate problem.

CendioOssman:
Can you share the entire output from rdesktop with us? It's could be a duplicate of #320.

Billybangleballs:
I'm sure the only output was the error message above, but I am unable to check this because I have downgraded back to a version that works.

CendioOssman:
It will be difficult for us to debug this in that case. I'll close this for now and let's hope fixing #320 will also fix the issue you've seen.

bernhardu:
There got Debian bug [934897](https://bugs.debian.org/934897) opened and I tried to look into it.
I guess it might describe the same issue as in this issue.
Attached is a file containing some more information and a backtrace of that location and a
git bisect pointing to commit 82fce18, but that might have just uncovered it.
There is also a patch avoiding the issue, if the end of the buffer is already reached.
Can you confirm my findings? @CendioOssman

CendioOssman:
I can confirm that the issue was probably already there, just hidden since we didn't have any sanity checks.

I can't say if the patch is correct or not. Have you looked at Microsoft's protocol documentation and see what it says about this?

bernhardu:
So does this justify to reopen this issue?

In version 50 or 35 (unfortunately no older version available as far as I see)
this section seems related:
[...]

Issue #320, opened 2019-03-29
"next order pointer would overrun stream" error connecting to Windows Server 2003 R2


opoplawski:
One of our users is reporting the following crash:

```
Connection established using SSL.
Protocol(error): rdp_protocol_error(), process_secondary_order(), next order pointer would overrun stream
0000 ff ff 18 04 05 10 01 40 01 00 10 59 0d 38 10 01 .......@...Y.8..
0010 cc 06 00 00 00 00 00 00 00 00 00 00 00 00 00 00 ................
0020 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 ................
```
[on for pages....]

This message seems to have come in with the recent security fixes: https://github.com/rdesktop/rdesktop/commit/4dca546d04321a610c1835010b5dad85163b65e1

opoplawski:
This does not appear to always happen - I can connect fine.

CendioOssman:
Do you think you could work with the user to figure out when this happens? It will be difficult to fix without being able to reproduce it.

sircrimer:
I get the following error 100% of the time after trying to establish a connection to a Windows 2000 Server. It works fine with a Windows 2003 Server.

`WARNING: Remote desktop does not support colour depth 24; falling back to 8
ERROR: rdp_protocol_error(), process_secondary_order(), next order pointer would overrun stream0000 ff ff 18 04 05 10 01 40 01 00 10 59 0d 38 10 01 .......@...Y.8..
0010 cc 04 28 00 40 04 28 00 40 04 28 00 40 04 28 00 ..(.@.(.@.(.@.(.
0020 80 04 28 00 00 04 28 00 00 04 28 00 00 04 28 00 ..(...(...(...(.
0030 80 04 28 00 80 04 28 00 80 04 28 00 40 04 28 00 ..(...(...(.@.(.
0040 40 04 28 00 40 04 28 00 40 04 28 00 40 04 28 00 @.(.@.(.@.(.@.(.`

CendioOssman:
Not sure, I'll have to inspect things closer.

Unfortunately I have to say that this will not be prioritized very highly. Windows 2000 is extremely ancient and our focus is on current versions of Windows.

grubba:
I seem to have this same issue with connecting to a Windows 2003 server.
It appears that the display window size affects whether it crashes or not.
My display is 3840x2160 pixels.
`rdesktop -g 90% host` crashes with
```
[...]

PR #346, opened 2019-09-07
Windows XP: Don't expect a userid on a PDU that doesn't have one

malxau:
Disclaimers:

1. Note this is against 1.8.x, since master has many other problems talking to older systems.
2. I don't know what I'm doing.
3. This doesn't appear to conform to the documentation, but does conform to the implementation.  If a user is logged onto an XP system and subsequently attempts to remote desktop into that system, the client receives a message type 6 that looks like this:
```
RDP packet #6, (type 6)
0000 04 00 16 00                                     ....
```
which doesn't conform to the TS_SHARECONTROLHEADER in that it's only four bytes, not six.

Without this, it is possible to use rdesktop 1.8.6 against XP, but not if the user has already logged on thereby implying a resolution change to an existing session.  1.8.2 works correctly.

### DIFF
--- a/rdp.c
+++ b/rdp.c
@@ -137,13 +137,17 @@ rdp_recv(uint8 * type)
 		return rdp_s;
 	}
 	in_uint16_le(rdp_s, pdu_type);
-	in_uint8s(rdp_s, 2);	/* userid */
 	*type = pdu_type & 0xf;
 
 #if WITH_DEBUG
 	DEBUG(("RDP packet #%d, (type %x)\n", ++g_packetno, *type));
 	hexdump(rdp_s->data + g_next_packet, length);
 #endif /*  */
+
+	if (*type != RDP_PDU_DEACTIVATE)
+	{
+		in_uint8s(rdp_s, 2);	/* userid */
+	}
 
 	g_next_packet += length;
 


CendioOssman:
Thanks, but this would break servers that send a correct packet. It would need to be more tolerant.

The hope is to get 1.9.0 out very soon instead of a 1.8.7. Could you test the patch on #342 and see if that helps? What other failures are you seeing with XP? Is it #320?

malxau:
Firstly, thank you for providing your patch.  I see why you want it to be more tolerant and I see how your patch achieves that.

Second, thank you for connecting me with @bernhardu on #342 .  It looks like we were seeing the exact same thing.  The issues I've had on older systems aren't related to #320, whose bug was already fixed; that just became a place to discuss that master doesn't work with older systems.  @bernhardu correctly pointed out this is caused by commit 90fd660803282713c3e2d29207dc26255f2e3899 (which I hadn't been able to diagnose) and after reverting that commit, I can connect to NT 4, 2000, XP, and 2003 from master.

I tested the patch on #342 and it does address the problem of connecting to a previously existing XP session.

Just some questions about that patch to make sure your choices were intentional though:

- If the input packet contains 3 bytes, this will try to overread, detect the condition, and drop the connection.  I assume that's intentional.
- If the input packet contains 4 bytes, this will log the condition and continue.
- If the input packet contains 5 bytes, this will try to overread and drop the connection (is this intentional?)

This means when calculating the length at the end the only time it is set to zero is when it started as exactly 4.

CendioOssman:
> Second, thank you for connecting me with @bernhardu on #342 . It looks like we were seeing the exact same thing. The issues I've had on older systems aren't related to #320, whose bug was already fixed; that just became a place to discuss that master doesn't work with older systems. @bernhardu correctly pointed out this is caused by commit [90fd660](https://github.com/rdesktop/rdesktop/commit/90fd660803282713c3e2d29207dc26255f2e3899) (which I hadn't been able to diagnose) and after reverting that commit, I can connect to NT 4, 2000, XP, and 2003 from master.

Great. I've opened #348 for trying to sort that out. Please continue testing there.



> Just some questions about that patch to make sure your choices were intentional though:
> 
> * If the input packet contains 3 bytes, this will try to overread, detect the condition, and drop the connection.  I assume that's intentional.
> 
> * If the input packet contains 4 bytes, this will log the condition and continue.
> 
> * If the input packet contains 5 bytes, this will try to overread and drop the connection (is this intentional?)
> 
> 
> This means when calculating the length at the end the only time it is set to zero is when it started as exactly 4.

Yes. This is non-standard behaviour so I want to be very specific about what we tolerate. A shorter package means we don't know if the PDU type is correct. A longer packet and we don't know if it is a cropped PDU source, or part of the data. So we'll deal with those if they are encountered in the wild.

PR #347, opened 2019-09-07
Windows 2000: Give effect to the comment about a field that can be negative and must be treated as signed

malxau:
Disclaimers:

1. Note this is against 1.8.x, since master has many other problems talking to older systems.
2. I don't know what I'm doing.

There's a comment at the top of process_secondary_order that indicates that the length field must be treated as signed.  (Yeah, I know.)  See also PR #190 where another developer looked this up in the specification.  The security changes (understandably) didn't preserve this behavior, which causes problems when talking to Windows 2000 servers.

### DIFF
--- a/orders.c
+++ b/orders.c
@@ -1250,7 +1250,7 @@ process_secondary_order(STREAM s)
 	/* The length isn't calculated correctly by the server.
 	 * For very compact orders the length becomes negative
 	 * so a signed integer must be used. */
-	uint16 length;
+	sint16 length;
 	uint16 flags;
 	uint8 type;
 	size_t next_order;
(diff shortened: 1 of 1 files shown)


malxau:
I saw that master had a different fix for this, so I'm replacing mine with a cherry-pick of that.  When reading the change, note that length starts as negative but after the +7 becomes positive when talking to Windows 2000.  s_check_rem is going to cast the signed value to unsigned, so if it's still negative at that point and the incoming packet didn't have enough contents for the unsigned value, the packet is about to be rejected.

One thing that scares me a little with this change is that s_check_rem casts to unsigned but the calculation for next_order is adding a signed value to an unsigned value, so if the packet were large enough it becomes possible to pass the s_check_rem check but then set the next offset to consume backwards.  I'm not sure if this is what the specification intended (but I'm not sure it's not either.)

CendioOssman:
The goal right now is to get 1.9.0 out instead of a 1.8.7, so if this fix is already on `master` then I'm afraid we consider this fixed. Please try to see what issue is preventing `master` from working with those older systems instead.

Issue #356, opened 2019-11-09
Segfault after "Retrying with plain RDP"


bavier:
```
$ rdesktop -h 2>&1 | head -n2
rdesktop: A Remote Desktop Protocol client.
Version 1.9.0. Copyright (C) 1999-2016 Matthew Chapman et al.
$ rdesktop -v ...
Autoselecting keyboard map 'en-us' from locale
is_wm_active(): WM name: GNOM
Connecting to server using SSL...
Protocol(warning): Protocol negotiation failed with reason: SSL not allowed by server
Retrying with plain RDP.
Segmentation fault
```
The target server seems to support only RDP version 4, so connecting with `-4` works, but I would expect a graceful fallback versus a segfault.

CendioOssman:
Indeed. Do you think you could try running it through `gdb` so we can get a backtrace?

bavier:
The backtrace I got from the system package does not look very enlightening:

```
$ gdb /gnu/store/j4x8jkhw2vdnllwb09wrz4c1zjm6ii93-rdesktop-1.9.0/bin/rdesktop
GNU gdb (GDB) 8.3.1
Copyright (C) 2019 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-unknown-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from /gnu/store/j4x8jkhw2vdnllwb09wrz4c1zjm6ii93-rdesktop-1.9.0/bin/rdesktop...
(No debugging symbols found in /gnu/store/j4x8jkhw2vdnllwb09wrz4c1zjm6ii93-rdesktop-1.9.0/bin/rdesktop)
(gdb) start ...
Function "main" not defined.
Make breakpoint pending on future shared library load? (y or [n]) y
Temporary breakpoint 1 (main) pending.
[...]

PR #357, opened 2019-12-02
use correct modulus and exponent in rdssl_rkey_get_exp_mod

markusbeth:
fixes issue #356


markusbeth:
I also ran into the issue described in #356. This is the patch that fixed it for me.

CendioOssman:
Nice find. Thank you for the fix. Sorry about taking so long to look at it.

Issue #349, opened 2019-09-27
How make cert-ignore?


dronmaxman:
I have use rdesktop 1.9.0 with vmware horizon. And I can't press YES.
What key help me to skip server certificate cheack?

![Screen Shot 2019-09-28 at 02 28 30](https://user-images.githubusercontent.com/15871669/65807641-154a9a00-e198-11e9-8a9a-6695008a4e8a.jpg)

CendioOssman:
I'm afraid we have not added a command line argument to ignore bad certificates. So right now you need to run rdesktop at least once to approve that certificate, or make sure the certificate is otherwise trusted by the system.

offsides:
Can I also request a command-line switch to auto-accept certs?  I use multiple SSH tunnels to bounce my rdesktop connections, and have a script to do it all for me.  Having to manually establish the tunnels just to be able to accept the certificate from a known system (just because it auto-renews every 6 months, if not more often) is pretty frustrating.  I understand why you don't want to auto-accept by default, but not even having the option is unfortunate.

nathanshearer:
I would also like to request an option similar to the one that wget uses:
--no-check-certificate

For now, I have been piping "yes" into rdesktop, but I don't like blindly answering yes to whatever it is asking me for all my connections.

MarkLTZ:
Hello,
PR #364 should address this issue. Could someone check and merge it on master repo?

petersun1977:
> 
> 
> I'm afraid we have not added a command line argument to ignore bad certificates. So right now you need to run rdesktop at least once to approve that certificate, or make sure the certificate is otherwise trusted by the system.

I can accept type "rdesktop RDS_SERVER" every time the cert is changed(for us it is every 6 months), but doing this only adds a new cert in /home/USER/.local/share/rdesktop/certs/known_certs, but does not remove the old cert, so the user has to remove old one manually.

MarkLTZ:
@petersun1977 did you check my PR? #364

twoninenine:
@dronmaxman if this is still relevant for you, I fount out following solution
`echo 'yes' | rdesktop <ip>`

jctmf:
Need it too.

HomeBoy38:
Ignore certificate is definitely an option I will hope to see some day.

[...]

PR #364, opened 2020-03-03
Add '-I' option for ignoring certificate at startup

MarkLTZ:
This close issue #349

### DIFF
--- a/rdesktop.c
+++ b/rdesktop.c
@@ -100,6 +100,7 @@ RD_BOOL g_packet_encryption = True;
 RD_BOOL g_desktop_save = True;	/* desktop save order */
 RD_BOOL g_polygon_ellipse_orders = True;	/* polygon / ellipse orders */
 RD_BOOL g_fullscreen = False;
+RD_BOOL g_ignore_certificate = False;
 RD_BOOL g_grab_keyboard = True;
 RD_BOOL g_local_cursor = False;
 RD_BOOL g_hide_decorations = False;
@@ -181,6 +182,7 @@ usage(char *program)
 #ifdef WITH_SCARD
 	fprintf(stderr, "   -i: enables smartcard authentication, password is used as pin\n");
 #endif
+	fprintf(stderr, "   -I: ignore certificate\n");
 	fprintf(stderr, "   -f: full-screen mode\n");
 	fprintf(stderr, "   -b: force bitmap updates\n");
 	fprintf(stderr, "   -L: local codepage\n");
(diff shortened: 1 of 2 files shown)


CendioOssman:
This is a bad idea from a security point of view, so at the very least it should have large warnings. And perhaps require a long option.

ThomasAdam:
Hi @MarkLTZ,

What's the rationale for this change?  Whenever things like this come along, it's usually due to a problem elsewhere, and such a change would be wide-open to security risks.

MarkLTZ:
@ThomasAdam this change close issue #349 

In my company we used this solution for starting in a batch way rdesktop where nobody can answer "yes" to accept server certificate.

The "-l" option is not the default, so this not affect standard security.

MarkLTZ:
@CendioOssman are long options already implemented  in rdesktop?

CendioOssman:
No. But it already uses `getopt()`, so it should hopefully not be that difficult to add support.

jasonqpswilson:
If cant have the ignore option is there a process/script to update the cache of known certificates to pre-populate it with the certificates required? Tried using 'openssl s_client -connect server:3389' and extract certificate from there, but the certificate does not match what rdesktop loads into the the known_certs file.

I have same issue - cant use the input stream to say yes when also using the password from stdin option.
All our servers have their certificates renewed 3 monthly with a self-signed certificate and automation breaks when they are renewed.

Jason

dronmaxman:
I also need this option. I am using thinstation6 with vmware-view client to connect to Windows Server 2019.

oldlavender:
Just adding some comments, besides the **security concerns** already mentioned by @CendioOssman:

### 1) Docs
This patch (and multiple others, including merged ones) misses the inclusion of the option in the docs. I am currently revieweing it, and the man pages have serious problems about corresponding to actual switches in the code.
### 2) Patch for long options (to overcome the security concerns)
I am currently working on a patch to provide proper support for --long options, but it's not that simple. It's been implemented before and then removed in the same day at da3dcb8fbe7c02a1e17887296d71124dc0d47d71 for it broke rdesktop in other operating systems. It's non-POSIX and differ in implementation between operating systems, for instance, in FreeBSD's man pages for getopt_long state it's different from GNU implementation. 

I am still considering options (and I'd be glad to have any input) on how to properly implement that, as I do not have room for VMs to test it on other OSes and also have a very poor idea of which OSes would be important these days. I can curently contemplate taking a list of key OSes, looking at their docs for getopt_long (if they support it) and them implement it accordingly and then having some users to test on them. But I am also considering develop our own parsing for options, which would be way easier to implement, but would add something else to maintain. 

After solving that, if @MarkLTZ would like to rewrite this patch (or if Mark's not interested anyone or I can do it) according to this new structure that would be great.

Mailaender:
> This is a bad idea from a security point of view

Piping in `yes |` to workaround this is far worse.

nneul:
If the tool would just USE THE GRAPHICAL UI to prompt, it wouldn't be a problem. The issue is that if you start rdesktop it prompts on the _console of the command_ instead of in the UI that you're actually working with. Just pop up the prompt in the actual graphical session and ask for whether or not to accept the cert there. That would eliminate the need for the ugly hacks we're having to do.

PR #375, opened 2020-06-11
wrong string was being null terminated

khalilfazal:


khalilfazal:
the only error is from the Apple VM. it should be ignored.

ThomasAdam (changes requested):
Hi @khalilfazal,

Thanks for your PR, unfortunately, this diff contains a tonne of unrelated code indentation changes which makes it impossible to see the real change you're fixing.

Could you resubmit this, minus these changes so it can be reviewed?

Thanks.

khalilfazal:
Hi Thomas,

I've removed the whitespace changes

manfredsc:
This pull request is simply a mess, and on top of it, it is wrong. You changed the variable name
only at one place and forgot the sizeof() instance.

The correct fix seems to be
```
--- rdesktop-1.9.0/rdesktop.c.orig      2019-09-11 16:33:07.000000000 +0200
+++ rdesktop-1.9.0/rdesktop.c   2024-01-11 14:42:25.416099137 +0100
@@ -1986,7 +1986,7 @@ save_licence(unsigned char *data, int le
        path[sizeof(path) - 1] = '\0';

        snprintf(tmppath, PATH_MAX, "%s.new", path);
-       path[sizeof(path) - 1] = '\0';
+       tmppath[sizeof(tmppath) - 1] = '\0';

        fd = open(tmppath, O_WRONLY | O_CREAT | O_TRUNC, 0600);
        if (fd == -1)

```

PR #391, opened 2021-07-28
Add '-F' option to disable fullscreen toggle using Ctrl-Alt-Enter

jcpierri:
To avoid the situation where toggling full screen mode on/off using Ctrl-Alt-Enter is undesirable, the -F option can be used to disable the toggle.

### DIFF
--- a/rdesktop.c
+++ b/rdesktop.c
@@ -100,6 +100,7 @@ RD_BOOL g_packet_encryption = True;
 RD_BOOL g_desktop_save = True;	/* desktop save order */
 RD_BOOL g_polygon_ellipse_orders = True;	/* polygon / ellipse orders */
 RD_BOOL g_fullscreen = False;
+RD_BOOL g_fullscreen_toggle = True;  /* control full-screen toggle using CTRL-ALT-ENTER */
 RD_BOOL g_grab_keyboard = True;
 RD_BOOL g_local_cursor = False;
 RD_BOOL g_hide_decorations = False;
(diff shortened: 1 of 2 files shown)


jcpierri:
Perhaps I should have explained better why this is a usefull feature (for me, at least)

I'm using a kind of _soft_ fullscreen mode where window decorations are removed (-D) and window geometry precisely matches the screen. This is made in order to keep window manager's key bindings (-K) available when more than one session is connected, as all of them are  in this _soft_ fullscreen mode.

This works flawlessly under FVWM, however, it becomes an issue if the user accidentally toggle to real fullscreen mode (using Ctrl-Alt-Enter), because then his keystrokes will not reach FVWM anymore, leaving him stuck in that one session until fullscreen mode get toggled off.

To avoid that nuisance I created the -F option to disable the toggle using the Ctrl-Alt-Enter combination.

oldlavender:
Hello @jcpierri sorry you had to wait so long for an answer and thanks for submitting this patch!

I understand that this patch is intended to solve a problem of an accidental keypress for users with multiple sessions open using this idea of 'soft' fullscreen. 

What I am trying to understand deeply is how users are getting stuck, it's still somewhat unclear to me. I haven't used FVWM the last 10-15 years, but is it a case of conflicting key bindings? Could you please provide some extra input on that? 

Assuming the case is conflicting keys, I think your solution is a good start and something to include in the next 1.9.x release (I have created a patches branch for that). I also think we could work out some design improvement for the future too that does not require an extra switch specifically for suppressing a single key binding. It would be great to count on you for that too, if it's possible.

jcpierri:
Hi brunodOut !

As I said, the reason I used rdesktop in this "soft fullscreen" mode is exactly because I wanted all window manager's key bindings accessible, allowing the user to activate other local tools on top of rdesktop screen as required as well as to allow them switch between multiple "soft fullscreen" rdesktop sessions.
If the user accidentally (or not) switches to real fullscreen (by hitting Ctrl-Alt-Enter) nothing changes on their screens but they lose the ability to use the window manager's key bindings.
As rdesktop already have an option to make it enter in full screen (-f) I thought it would be useful to have another option to forbid fullscreen mode, and -F was born.

cyguard2:
Hello @[jcpierri](https://github.com/jcpierri), I found your patch very useful.
I want to completely avoid the user to exit the full screen mode with Ctrl+Alt+Enter, and this is what I was searching for. Thanks a lot.

PR #394, opened 2021-10-27
Fix segfault in utils_cert_handle_exception if util_dialog_choice

opoplawski:
returns NULL.

Downstream bugreport: https://bugzilla.redhat.com/show_bug.cgi?id=2008044

### DIFF
--- a/utils.c
+++ b/utils.c
@@ -1041,7 +1041,7 @@ utils_cert_handle_exception(gnutls_session_t session, unsigned int status,
 
 	/* show dialog */
 	response = util_dialog_choice(message, "no", "yes", NULL);
-	if (strcmp(response, "no") == 0 || response == NULL)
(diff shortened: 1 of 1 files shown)


brunodOut:
Hello, @opoplawski 

Thanks for reaching out and sorry it took a while to answer.

In a quick look I can easily see what's happening in the code level.

However, I get a Bug Access Denied message from bugzilla even after logged in. If you could share at least the steps to reproduce it that would be great.

Anyway I'll be merging it soon and test with some other patches that are pending.

oldlavender:
Just an update, yesterday I merged this patch into a new branch `patches` (I just created it for changes that need a more time pressing response, on the contrary to the `next` branch that I am also creating for new features).

Also, thanks for submitting this patch!

